PR #59, opened 2012-12-28
Fix double serialization for very large exponents that loose precision with %f printf format flag to %g format flag

birkler:
...h very large exponents without loosing precision.

### DIFF
--- a/tests/test_parse.c
+++ b/tests/test_parse.c
@@ -55,9 +55,19 @@ static void test_basic_parse()
 	printf("new_obj.to_string()=%s\n", json_object_to_json_string(new_obj));
 	json_object_put(new_obj);
 
-	new_obj = json_tokener_parse("12.3");
-	printf("new_obj.to_string()=%s\n", json_object_to_json_string(new_obj));
-	json_object_put(new_obj);
+	{
+		double testValue = 12.3;
+		json_object* temp_obj = json_object_new_double(testValue);
+		new_obj = json_tokener_parse(json_object_to_json_string(temp_obj));
+		double readValue = json_object_get_double(new_obj);
+		if (fabs(testValue - readValue) < 1e-10) {
+			printf("new_obj.to_string()=12.300000\n"); //What is expected by the script
+		} else {
+			printf("new_obj.to_string()=%s\n",json_object_to_json_string(new_obj));
+		}
+		json_object_put(new_obj);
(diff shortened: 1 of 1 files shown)


hawicz:
It looks like the approach from that paper will do what's needed here.  Or rather, the approach from a subsequent paper by Florian Loitsch titled "Printing floating-point numbers quickly and accurately with integers" (http://florian.loitsch.com/publications/dtoa-pldi2010.pdf?attredirects=0), known as the Grisu3 algorithm.
There's even an implementation of it at https://github.com/google/double-conversion, and I've checked that it works as expected on the few values mentioned above.  Unfortunately, it's C++ instead of C, so we can't just drop in a copy. :(

ploxiln:
A new algorithm, "Errol3" (and presumably minor update "Errol4") has been developed, which is apparently half the speed of "Grisu3", but always produces the optimal decimal string output (Grisu3 only does so for 99.5% of possible doubles).

And, there's a plain C implementation available from the authors, under the MIT license:
https://github.com/marcandrysco/Errol/tree/master/lib

paper: http://cseweb.ucsd.edu/~lerner/papers/fp-printing-popl16.pdf
explanatory discussion with links: https://news.ycombinator.com/item?id=10915182

MCausc78:
💀💀💀💀💀

hawicz:
I've been playing around with the Errol code, and while it seemed initially promising there are a fair number of gotcha's that make it difficult to use.  First of all, it just flat out fails on negative values and 0.0, with either buffer overruns or infinite loops.  That's not too hard to work around, but not particularly confidence inspiring.
Then, I noticed that `errol4_dtoa()` doesn't actually produce optimal output, turning double values like `12.3` into "12300000000000001" exponent 2.  `errol3_dtoa()` seems to get that right, but again, not confidence inspiring.
Finally, we'd need to write some code to handle inserting the decimal point, adding extra zeros in various cases (e.g. double `10` => "1" exponent 2, needs a "0" appended), and deciding whether instead to append `e<exponent>`.

As an alternative, I found `g_fmt()` from https://netlib.org/fp/ which, with just slight tweaks (see https://github.com/hawicz/json-c/tree/shorter-float-output), handles all of this.  I haven't had a chance to benchmark it, so I don't yet know whether the netlib code will perform well enough to include in mainline json-c.

PR #61, opened 2013-01-08
Make json_object_object_add() indicate success or failure, test fix

TheCount:
json_object_object_add() now indicates success or failure via its return value. The previous behaviour, namely to exit the program on failure, is cumbersome in large-scale applications.

Note that the change of return value from void to int may break the ABI so that programs using json-c may have to be rebuilt.

Unrelated to this change, I found I had to adjust the expected output of test_parse for a successful build.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -371,7 +371,7 @@ struct lh_table* json_object_get_object(struct json_object *jso)
   }
 }
 
-void json_object_object_add(struct json_object* jso, const char *key,
+int json_object_object_add(struct json_object* jso, const char *key,
 			    struct json_object *val)
 {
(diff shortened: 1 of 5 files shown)


hawicz:
I'm mostly convinced that changing the return value from void to int won't change the ABI (see http://stackoverflow.com/questions/15626579/c-abi-is-it-safe-to-change-void-function-to-return-int) so this change should be merged.  At this point, I'll need to merge it by hand.

The fix to test_parse has already been applied.

hawicz:
ugh, there's is some pretty bogus use of goto's in this change. :(

hawicz:
After dealing with the steaming pile that is git, I finally managed to get the change to json_object_object_add manually merged in.

PR #113, opened 2013-11-18
Build fixes (make dist and make distcheck)

rossburton:
A few fixes to reduce warnings from autoreconf, fix "make dist", make "make distcheck" compile (but not pass, yet) and remove generated files from git.


hawicz:
The change to use top_srcdir was fine; I didn't realize at the time of my previous comment that the AM_CPPFLAGS setting is what would affect all of the other test programs.
There were a number of other things that needed to be fixed, but I believe I've now addressed those so a "make distcheck" now completes without error.
I don't think the tarball created by "make dist" is entirely complete as a distribution tarball, but that's not the process followed for releases (yet, anyway), so fixing that is a separate issue.

PR #155, opened 2014-08-21
add bsearch for arrays

LeSpocky:
Arrays can already be sorted with json_object_array_sort() which uses
qsort() of the standard C library. This adds a counterpart using the
bsearch() from C.


hawicz:
Sounds good.  I'll merge it, but if you could add some tests for this, I would appreciate it.

PR #162, opened 2014-12-05
json_util: add directory check for POSIX distros

commodo:
On a Linux based target (OpenWRT to be more explicit), a directory path
to json_object_from_file() returns a non-NULL object.

On WIN32 the docs say:
  http://msdn.microsoft.com/en-us/library/z0kc8e3z.aspx

EACCES
    Tried to open a read-only file for writing, file's sharing mode does
    not allow the specified operations, or the given path is a directory.

On Linux/Unix:
  http://linux.die.net/man/2/open

EISDIR
    pathname refers to a directory and the access requested involved
    writing (that is, O_WRONLY or O_RDWR is set).

The quickest/easiest fix would be to change O_RDONLY to O_RDWR.

I feel it's safer to explicitly check on POSIX targets if the file
is a directory.

I would be fine with whatever approach, as long as it is reviewed
by the creator/maintainer.

### DIFF
--- a/json_util.c
+++ b/json_util.c
@@ -71,6 +71,20 @@ struct json_object* json_object_from_file(const char *filename)
   char buf[JSON_FILE_BUF_SIZE];
   int fd, ret;
 
+#ifdef HAVE_SYS_STAT_H
+  struct stat st;
+
+  if (stat(filename, &st) < 0) {
+    MC_ERROR("json_object_from_file: could not stat file %s: %s\n",
+	     filename, strerror(errno));
+    return NULL;
+  }
+  if ((st.st_mode & S_IFMT) == S_IFDIR) {
+    MC_ERROR("json_object_from_file: path is a directory %s: %s\n",
+	     filename, strerror(errno));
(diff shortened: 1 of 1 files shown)


hawicz:
How do you get the contents of a directory to parse as a valid JSON object?  When I try it, I get NULL because parsing the raw directory bytes fails.
Even if this really is something that needs to be checked for, it should be done with a fstat call after the file is open.  Also, the code was reorganized to split off a json_object_from_fd function, so your patch would need to be re-worked anyway.

commodo:
Well, the issue was some time ago, and in our code I'm checking if the filename is a folder before reading it.

On OpenWRT things are a bit different.
Filesystems are different, kernel is usually the same (some things are patched differently), and uClibc or musl is used instead of glibc.
Don't know what causes the issue, but for now I'll leave it as is.
A lot of updates happened on OpenWRT since I submitted the PR, and this might not be required anymore, but I'll check later and if still reproduces, I'll create a new PR.

PR #174, opened 2015-03-23
Adding JSON_C_TO_STRING_PRETTY_TAB flag

haata:
- Tabs are easier to read for tired eyes and editor adjustable

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -300,7 +300,14 @@ static void indent(struct printbuf *pb, int level, int flags)
 {
 	if (flags & JSON_C_TO_STRING_PRETTY)
 	{
-		printbuf_memset(pb, -1, ' ', level * 2);
+		if (flags & JSON_C_TO_STRING_PRETTY_TAB)
+		{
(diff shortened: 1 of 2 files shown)


hawicz:
This could be simpler by having JSON_C_TO_STRING_PRETTY_TAB be a modifier of JSON_C_STRING_PRETTY.  It *is* defined as a flag after all, and could simply be or'd together.

haata:
I'd defined it as another bit. Would you prefer another argument instead for specialization?

hawicz:
No, I'm saying you can avoid most of the modifications to the code if you actually use it as a flag to augment JSON_C_STRING_PRETTY.  i.e.

    res = json_object_to_json_string_ext(jso, JSON_C_STRING_PRETTY|JSON_C_STRING_PRETTY_TAB)

haata:
@hawicz 
Sorry about the delay, work had me tied up.

Ah, yep, makes perfect sense. Updating the commit.

PR #179, opened 2015-04-02
Added array_list_del_idx and json_object_array_del_idx

Protovision:
array_list_del_idx lets you delete a number of elements from an array list starting at a specified index. json_object_array_del_idx uses array_list_del_idx to do the same thing for json object arrays.


hawicz:
Can we get some tests for this?

PR #184, opened 2015-05-20
Use only GCC-specific flags when compiling with GCC

dago:
Hi,

these are some adjustments on the autoconf-foo that enhance the compatibility with non-GCC compilers.

Best regards -- Dago

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,3 @@
-include Makefile.am.inc
-
 EXTRA_DIST = README.html README-WIN32.html config.h.win32 doc json-c.vcproj
 SUBDIRS = . tests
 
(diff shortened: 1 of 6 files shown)


hawicz:
Why did you drop AC_C_CONST and AC_TYPE_SIZE_T?

dago:
I only moved the two lines upwards, I can't remember the reason any more, probably because the manual suggested to have these tests early and before the header checks.

hawicz:
The use of gcc specific flags on non-gcc compilers should be fixed by changeset 80c1f69b9ef88ee0305776478b6207a7fe11340b
For anything else that's needed, please submit a separate pull request.  However, munging the scripts to support lackluster shells leaves a bad taste in my mouth.  If you can instead arrange for the tests to be executed with a fully working shell I'd be much happier to merge the change.

PR #191, opened 2015-08-21
Override int64 to only display uint64 strings

haata:
- Not true uint64 support, but should handle all cases properly
- Useful for when dealing with large 64 bit numbers without negatives

I've also done some initial code to implement true uint64 support but it's quite invasive (don't have all test cases passing yet either). The project I'm using json-c with doesn't not require signed numbers at all but it does deal with extremely large 64 bit numbers.

I'm not terribly worried if there is push back from this pull request.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -509,6 +509,8 @@ static int json_object_int_to_json_string(struct json_object* jso,
 					  int level,
 					  int flags)
 {
+	if (flags & JSON_C_TO_STRING_UINT64)
+		return sprintbuf(pb, "%"PRIu64, jso->o.c_int64);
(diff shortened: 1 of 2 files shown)


hawicz:
ew... kind of gross, but effective I suppose.  Please rename the flag to JSON_C_TO_STRING_UNSIGNED, and add at least one test, and I'll merge it.
However, this will create output that json-c can't parse, right?

hawicz:
We now have a `json_object_new_uint64` and json_tokener will create nodes of type json_object_int_type_uint64` when the value parsed is greater than INT64_MAX, so I don't think this is needed anymore.

PR #193, opened 2015-09-03
Make serialization format of doubles configurable

### LINKED ISSUE #154: double precision
When double object are converted to string, there is often a lost of precision (see json_object_double_to_json_string).

I suggest doing something like that, to avoid conversion noise :
size = snprintf(buf, sizeof(buf), "%.*g",DBL_DIG-1, jso->o.c_double);

jplitza:
Using the new function `json_object_new_double_fmt(double d, char *fmt)`, one can now specify what printf-compatible format string should be used to serialize this specific double. This allows for dynamic adaptation of the precision depending on the precision of the source as well as the indended usage.

This fixes #154 and #59.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -55,7 +55,6 @@ static struct json_object* json_object_new(enum json_type o_type);
 static json_object_to_json_string_fn json_object_object_to_json_string;
 static json_object_to_json_string_fn json_object_boolean_to_json_string;
 static json_object_to_json_string_fn json_object_int_to_json_string;
-static json_object_to_json_string_fn json_object_double_to_json_string;
 static json_object_to_json_string_fn json_object_string_to_json_string;
 static json_object_to_json_string_fn json_object_array_to_json_string;
 
@@ -644,10 +643,10 @@ int64_t json_object_get_int64(const struct json_object *jso)
 
 /* json_object_double */
 
-static int json_object_double_to_json_string(struct json_object* jso,
-					     struct printbuf *pb,
-					     int level,
-						 int flags)
+int json_object_double_to_json_string(struct json_object* jso,
+				      struct printbuf *pb,
+				      int level,
+				      int flags)
 {
   char buf[128], *p, *q;
   int size;
@@ -663,7 +662,8 @@ static int json_object_double_to_json_string(struct json_object* jso,
     else
       size = snprintf(buf, sizeof(buf), "-Infinity");
   else
-    size = snprintf(buf, sizeof(buf), "%.17g", jso->o.c_double);
+    size = snprintf(buf, sizeof(buf),
+        jso->_userdata ? (const char*) jso->_userdata : "%.17g", jso->o.c_double);
 
   p = strchr(buf, ',');
   if (p) {
--- a/json_object.h
+++ b/json_object.h
@@ -614,6 +614,9 @@ extern int64_t json_object_get_int64(const struct json_object *obj);
 /* double type methods */
 
 /** Create a new empty json_object of type json_type_double
+ *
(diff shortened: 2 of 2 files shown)


jplitza:
I rebased the change to reflect current master status. However, I would really like to have some feedback. Is this attempt of a solution for various serialization problems regarding doubles welcome? Is it in the completely wrong direction? Should I abandon all hope of it ever getting merged, or is there something about the implementation that I should change?

hawicz:
Using _userdata to store the format string is an abuse of the fields added to support json_object_set_serializer(), and will break anything that uses that.
Do you actually want a *per-object* format, or just a per-json_object_to_json_string()-call format?
If the latter, it may be better to extend json_object_to_json_string_ext() and add a new json_object_to_json_string_ctx() function that would allow some kind of additional information to be passed to customize the serialization process.

(note that I'm not suggesting a global setting to set the format for doubles, since adding such a thing would introduce the first bit configuration into the library that isn't specific to a particular json-object)

jplitza:
I am pretty sure that at the time of writing, I reassured myself that the "original" purpose of _userdata wasn't broken by this, but I may be wrong. And as the parameter was generically called _userdata, I assumed it to be fair to use it for this purpose.

I think such a format ideally has to be per-object, because different objects in the same JSON structure can very well have different number of significant digits (which is probably actually all that needs to be specified by the caller). But sure, a parameter for the whole string generation would be better than nothing.

hawicz:
I interpret "_userdata" as "data provided by the user", not as a field that can be re-used by the rest of the library for an internal (i.e. non-user-provided) serialization routine.
Do you actually have a specific use case where some doubles need one format string while others need a different one?  How do you know how many significant digits any particular one needs, to be able to pass the right things to json_object_new_double_fmt()?

I noticed that, although not as convenient as the json_object_new_double_fmt() function, this could be handled by exporting json_object_double_to_json_string(), changing the snprintf in there to say:

    size = snprintf(buf, sizeof(buf),
        jso->_userdata ? (const char*) jso->_userdata : "%.17g", jso->o.c_double);

and then suggesting in the docs for json_object_new_double() and/or json_object_double_to_json_string() that custom double output can be selected by calling:

    json_object_set_serializer(jso, json_object_double_to_json_string, "%.20g", NULL);

(no strdup and no delete_fn needed if the format is a string constant)

Anyway, after taking another look at your changes, if you still want a json_object_new_double_fmt() function I think it would be workable with a few changes:

1. Rename the _userdata to _serializer_data, and _user_delete to _serializer_delete
2. Do something to fix the memory leak introduced by strdup'ing the fmt arg to json_object_new_double_fmt().  Either set _serializer_delete to json_object_free_userdata, or drop he strdup and document that the fmt parameter must be a string constant or otherwise have a lifetime at least as long as the json_object being created.
3. Document that json_object_new_double_fmt is equivalent to a regular json_type_double with json_object_set_serializer(j, json_object_double_to_json_string, strdup(fmt), json_object_free_userdata) called afterwards.  (or perhaps, actually *do* that)
4. Add tests

jplitza:
Your approach with `json_object_set_serializer` totally makes sense, because it is kind of is exactly what that mechanism is intended for. I'll change that!

Regarding the example: We use the library to collect system information, where on the one hand we have data like loadavg (comes in with a precision of only two digits after the comma, but can be blown up to 0.2800000000000001 by the conversion), and on the other hand WGS84 coordinates, which would suffer severe precision loss if truncated to two digits after the comma. Surely, the conversion blowup is a problem that could also be addressed otherwise, but this seemed to be the easiest and most flexible way.

PR #194, opened 2015-09-08
Add utility function for comparing json_objects

hschaa:


rgerhards:
+1 for merging this

hschaa:
Ping, any update on this pull request?

hawicz:
Thanks, this looks like a useful thing to add, however:

* json_hash_equal isn't a great name, it's not actually calculating a hash of the two objects and comparing them, as the name might imply.  Can you please rename it to something like json_object_all_values_equal() ?

* Can you please use 0/1 instead of FALSE/TRUE?  I'd actually like to eliminate those defines.

* You don't define what "equal" means.  For example, are these two objects equals?:

    {"foo":123,"bar": 222}
    {"bar":222,"foo":123}

    If no, then json_hash_equal doesn't properly distinguish those and needs to be extended/fixed.
Either way, can you please flesh out the comment on json_object_equal() to better describe what it is implementing?

* The second loop in json_hash_equals doesn't need to call json_object_equal, because the first loop is guaranteed to have done that already for keys that exist in both objects.  It *only* needs to identify keys in jso2 that aren't present in jso1.

* Finally, can you please add some tests to verify that the code works?

hschaa:
Sure, I'll work on the comments and respin the PR as soon as possible.

PR #198, opened 2015-10-01
Fix possible memory leak and remove superfluous NULL checks before free()

unmole:


hawicz:
This change is either incomplete, or wrong.  If setlocale(LC_NUMERIC, NULL) can never return NULL, then the use of tmplocale is superfluous and can be removed.  On the other hand, if that *can* return NULL, then it's possible for oldlocale to be NULL, so the NULL checks are required.

unmole:
@hawicz If the pointer passed to free() is NULL, no action occurs. It is not necessary to check if a pointer is NULL before calling free() on it.

hawicz:
Sory, but json-c is used on many different systems and on some it *is* necessary to check for null.  Just because you can get away with it on your particular OS with your particular malloc implementation is not a reason to break it elsewhere.

rgerhards:
@hawicz just out of curiosity: which currently existing system is so broken that it is necessary to check for NULL when calling free? It's long standardized that this is a no-op, see http://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf section 7.20.3.2.

hawicz:
hmm... it seems that I'm wrong about calling free(NULL) being a problem.  It still feels wrong to me, but perhaps that just me getting old and crufty.  In some ways I like the fact that a check for null helps imply what's values are expected to be present, but now not enough to block this change.

PR #211, opened 2015-11-27
Atomic updates for the refcount

EmielBruijntjes:
I've modified the json_object_get() and json_object_put() functions to use atomic updates of the refcount. This allows json objects to be safely moved and copied between threads.

It uses the atomic builtin functions of GCC (see https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Atomic-Builtins.html). These functions are also already used inside the linkhash.c file of this json-c library, so I think that it is not much of a problem that they will be used inside json_object.c too.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -161,11 +161,13 @@ static int json_escape_str(struct printbuf *pb, const char *str, int len)
 extern struct json_object* json_object_get(struct json_object *jso)
 {
 	if (!jso) return jso;
+
 #if defined __GNUC__
     __sync_add_and_fetch(&jso->_ref_count, 1);
 #else
     ++jso->_ref_count;
 #endif        
+
 	return jso;
 }
 
--- a/json_object.c
+++ b/json_object.c
@@ -160,25 +160,29 @@ static int json_escape_str(struct printbuf *pb, const char *str, int len)
 
 extern struct json_object* json_object_get(struct json_object *jso)
 {
-	if (jso)
-		jso->_ref_count++;
+	if (!jso) return jso;
+#if defined __GNUC__
+    __sync_add_and_fetch(&jso->_ref_count, 1);
+#else
+    ++jso->_ref_count;
+#endif        
 	return jso;
 }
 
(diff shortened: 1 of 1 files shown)


hawicz:
ugh, no.  If you want to do something like this, please do not litter the code with a bunch of ifdefs.  Instead, define a macro (or inline function) in a header file to keep the compiler-specific cruft isolated.
If you clean up the existing usage in linkhash.c, that would be appreciated as well.

EmielBruijntjes:
Wow, that is such a nice and great response - given the fact that I not only found but also immediately solved a real bug, and did also my very best to keep the coding style as close to the rest of the json-c library as possible. I inspected the code to find a place in which atomic functions were already used, and I implemented it in _exactly the same way_. In fact, I almost copied that #ifdef block from the json-c source. Both the json_object.c and the linkhash.c files already are full of #ifdef statements. Considering all this, I find it strange that you respond to my change with "ugh".

My change basically comes down to two simple lines of code, it solves a genuine bug (refcounts in json objects are not thread safe) and it will probably take you about 1 minute and 20 seconds to convert these two lines into a format that you appreciate. The alternative is that we now start a discussion that will take many hours of both our lives before we have these two lines of code exactly formatted he way you like it.

hawicz:
Thanks for considering the existing code, and submitting a fix for a problem, but the fact is that what's already present is by no means perfect and can't always be used as an example without considering the change on its own merits.  Also, there's a bit of a difference between things like the REFCOUNT_DEBUG ifdefs in json_object.c, that serve to conditionalize a "feature", and those that work around platform inconsistencies, so IMO it's not valid to use that as an excuse for adding more ifdefs.
I suggested a reasonable alternative, and you're free to re-do your change or not, as you see fit, but do not presume to have a claim on the time that I spend on json-c.  I volunteer my time, and that "1 minute and 20 second" that you so flippantly claim it would take to adjust your changes is time that I could just as happily spend doing something else.

EmielBruijntjes:
I have posted it as issue instead.

PR #220, opened 2016-01-13
Add utility function for comparing json_objects

hschaa:


hschaa:
This should address all comments from the previous pull request #194

PR #231, opened 2016-05-29
Fix size_t size check for first-time builds

NeoRaider:
json_inttypes.h includes json_config.h, which is not available before
configure has run successfully at least once. Change configure check to
include stddef.h instead, which defines size_t.


NeoRaider:
Ah, sorry, I didn't notice that there's already PR #230 making the same change.

PR #232, opened 2016-05-29
tests/tests1: fix printf format for size_t arguments

NeoRaider:
Change %d to %zu for size_t arguments, otherwise compilation will fail with
errors like:

test1.c:70:15: error: format ‘%d’ expects argument of type ‘int’, but argument 2 has type ‘size_t {aka long unsigned int}’ [-Werror=format=]


hawicz:
Good idea, but unfortunately I don't think that'll work on Windows, at least apparently with anything older than VS2015.  I'd say let's use %llu and cast to unsigned long long.

NeoRaider:
I've updated the PR as suggested.

PR #241, opened 2016-07-20
getting error in date string 19\/07\/2016, fixed for error 19/07/2016

scbonde:
Hello,

While using json-c library, whenever date has to be sent as a string, library creates some invalid "\\/" string within date.

See below example :

Before changing JSON library code :
{
	"String Node" : "19\/07\/2016"
}

After changing JSON library code :
{
	"String Node" : "19/07/2016"
}

Please accept this pull request so that other's will NOT face this issue.

Thanks.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -139,7 +139,7 @@ static int json_escape_str(struct printbuf *pb, const char *str, int len, int fl
 			else if(c == '\f') printbuf_memappend(pb, "\\f", 2);
 			else if(c == '"') printbuf_memappend(pb, "\\\"", 2);
 			else if(c == '\\') printbuf_memappend(pb, "\\\\", 2);
-			else if(c == '/') printbuf_memappend(pb, "\\/", 2);
+			else if(c == '/') printbuf_memappend(pb, "/", 1);
 
 			start_offset = ++pos;
 			break;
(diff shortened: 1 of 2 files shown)


hawicz:
The backslashes are *not* invalid, and I will not be merging this.  See the definition of a string over at http://json.org
If you want to avoid them you can pass the JSON_C_TO_STRING_NOSLASHESCAPE flag.

scbonde:
Its okay, please suggest a way to print date in JSON string as mentioned like "19/07/2016".
Using "/" as a date separater. I don't want to use another separater that's only the limitation.
("19\/07\/2016" is invalid for date validation.)

hawicz:
Unless you're doubling the backslashes, their presence in the JSON string is irrelevant because once you parse it that value (as returned by json_object_get_string()) *will* be "19/07/2016".
If you're need to send the output from json-c to some other non-conforming JSON parser, or if you simply don't like seeing the extra backslashes in the output, you can pass the above-mentioned flag.

scbonde:
Yes, i have tried with above flag, in parsing there is no issue, but i need to create this date string and pass it to other JSON based device. At that time, i am troubled with validating date string.

PR #243, opened 2016-07-27
Fix #165

### LINKED ISSUE #165: json_object_to_json_string_ext length
To avoid needing a `strlen()` after calling `json_object_to_json_string_ext`, it would be very handy if there was a variant of the function which produced the length as well as the string.

jobol:


hawicz:
I'm not particularly thrilled about the name, but I can't think of a better one, so ok I'll merge this.

However, do you have any tests for this?  Adding a new feature without anything in the tests that actually checks that it works isn't so great.

jobol:
Maybe you prefer 'json_object_to_json_string_len' ?
I can switch.
However about test I'll do it quickly.

jobol:
done in a way that may hurt but ...

PR #248, opened 2016-08-07
Minor changes in C source code

RyDroid:


hawicz:
I'm not particularly thrilled about all these null checks.  I can see some value to being a bit forgiving of passing NULL to public API functions, but that lack of rigor shouldn't propagate to the internals of json-c.

I have not had a chance to look at the details of where all of these changed functions are called from, but given that many of them are static functions I'm a bit reluctant to merge this as-is.

RyDroid:
I removed the null checks in static functions.

json_object.c
```diff
@@ -167,8 +169,11 @@ static int json_escape_str(struct printbuf *pb, const char *str, int len, int fl
 extern struct json_object* json_object_get(struct json_object *jso)
 {
 	if (jso)
+	{
 		jso->_ref_count++;
-	return jso;
+		return jso;
+	}
+	return NULL;
```
kdopen: This gains nothing over the original code.

json_object.c
```diff
@@ -81,11 +81,12 @@ static void json_object_fini(void)
...
 		{
 			MC_DEBUG("json_object_fini: %d referenced objects at exit\n",
-			   json_object_table->count);
+				 json_object_table->count);
 			lh_foreach(json_object_table, ent)
 			{
 				struct json_object* obj = (struct json_object*)lh_entry_v(ent);
-				MC_DEBUG("\t%s:%p\n", json_type_to_name(obj->o_type), obj);
+				MC_DEBUG("\t%s:%p\n",
+					 json_type_to_name(obj->o_type), obj);
```
kdopen: It is generally preferable to keep formatting/whitespace changes in a separate commit from functional ones.

json_object.c
RyDroid: Without this commit, the value of jso was always read for the if and the return. With it, if with the first read it is NULL, it is not read a second time. Of course, that is the case for a stupid compiler that is not enough clever to optimize it-self.

json_object.c
RyDroid: I agree if the changes are not minor, but that is not the current case.

json_object.c
hawicz: I agree with Keith here.  That's really a mostly pointless nano-optimization.  With optimization completely turned off, the single extra memory access this saves will be totally swamped by everything else, and for a case that will rarely be hit.  When optimization is turned on, this change makes no difference at all (jso is both passed and returned in a register).
It's not enough to preclude me from merging the pull request, but please don't bother spending any time on changes like this in the future, unless they're part of a more comprehensive effort to speed things up, and you have actual measurements to back it up.

json_object.c
hawicz: Regardless of whether the changes are minor or not, if the only reason for changing a particular block of code is simply to adjust the formatting then that should go in a separate commit.
It's ok for now, since the changes are so small, but please keep them separate in the future.

hawicz:
I see that you still have NULL checks in the linkhash functions.  While they're not static, and the linkhash.h header does get installed, they should really be considered internal to json-c.
Is there some case where those checks are actually needed?  I would say that if that is the case, then that's a bug in the json-c code that needs to be fixed rather than papered over.

json_object.c
```diff
@@ -236,12 +241,15 @@ enum json_type json_object_get_type(const struct json_object *jso)
 }
 
 void* json_object_get_userdata(json_object *jso) {
-	return jso->_userdata;
+	return jso == NULL ? NULL : jso->_userdata;
```
kdopen: Largely a matter of style, but also compilers often optimize on the basis that the "true" part of a conditional is the most likely.

This would probably be more efficient if reversed, i.e.

    return jso ? jso->_userdata : NULL;

or even 

    return jso ? jso->_userdata : jso;

json_object.c
```diff
@@ -377,8 +385,7 @@ static int json_object_object_to_json_string(struct json_object* jso,
 	}
 	if (flags & JSON_C_TO_STRING_SPACED)
 		return sprintbuf(pb, /*{*/ " }");
-	else
-		return sprintbuf(pb, /*{*/ "}");
+	return sprintbuf(pb, /*{*/ "}");
```
kdopen: Why not

    return sprintbuf(pb,  /*{*/ (flags & JSON_C_TO_STRING_SPACED) ? " }" :  "}" );

and completely remove the `if`?

json_object.c
```diff
@@ -750,7 +755,7 @@ int json_object_userdata_to_json_string(struct json_object *jso,
 
 void json_object_free_userdata(struct json_object *jso, void *userdata)
 {
-	free(userdata);
+	if(userdata) free(userdata);
```
kdopen: `free` is NULL-safe

linkhash.c
```diff
@@ -685,5 +689,5 @@ int lh_table_delete(struct lh_table *t, const void *k)
 
 int lh_table_length(struct lh_table *t)
 {
-	return t->count;
+	return t == NULL ? -1 : t->count;
```
kdopen: Technically, this is a breaking change if this is a public method - but there are no comments in linkhask.h for this function.

I think the commit should add some (Doxygen) documentation in the header file to define this error condition.

json_object.c
```diff
@@ -510,7 +517,8 @@ json_bool json_object_object_get_ex(const struct json_object* jso, const char *k
 
 void json_object_object_del(struct json_object* jso, const char *key)
 {
-	lh_table_delete(jso->o.c_object, key);
+	if(jso)
+		lh_table_delete(jso->o.c_object, key);
```
kdopen: If you dig deeper, there may be a problem if `key` is NULL and the supplied hash functions are not NULL-safe due to this line

    unsigned long h = t->hash_fn(k);

in `lh_table_lookup_entry`

hawicz:
I'm closing this because we're not going to include these changes, since they're either minor style issues that aren't worth copying over by hand (or should not be done anyway), or no longer relevant given recent changes, or changes that would mask other bugs.
I did, however, make some changes to json_object_get_userdata and json_object_set_userdata inspired by what was suggested here, in changeset d4899bd.

PR #249, opened 2016-08-07
Improving README

RyDroid:


hawicz:
Please don't delete the README and NEWS files.  They need to be present to avoid warnings from automake.

RyDroid:
@hawicz Ok. I filled NEWS and README in order that this files do not seem to be an error.

PR #250, opened 2016-08-07
Improving .gitignore

RyDroid:


hawicz:
As in issue #251, this "license" notice seems odd.  Also, IMO it's not worth introducing another license just for some minor changes to the .gitignore file.

RyDroid:
I removed the license notice.

PR #251, opened 2016-08-07
Adding a file for EditorConfig

RyDroid:


hawicz:
That notice at the top, which doesn't include a copyright, is a bit odd.  Is there some reason that the standard MIT license doesn't apply to this file?

RyDroid:
There is no "MIT license".
https://www.gnu.org/licenses/license-list.en.html#X11License
https://www.gnu.org/licenses/license-list.en.html#Expat

Except, the license header, the file is smaller than what you call "the standard MIT license", so I think that this license is too big for such a file. Moreover, this license gives attribution, I don't want it for such a small thing.
It is a slightly modified version of the GNU All-Permissive License. https://www.gnu.org/licenses/license-list.en.html#GNUAllPermissive

hawicz:
In my experience as commonly used, and as I intended it, the phrase "MIT License" refers to a license like https://opensource.org/licenses/MIT, and matching what is in json-c's COPYING file.  

gnu.org, though a convenient reference, isn't authoritative for what is and what isn't a valid license or what it is called.  However, if you want to use their terminology, please interpret "MIT License" to mean "Expat License".

Whatever you call it is largely irrelevant though, as my complaints with the license in the .editorconfig file are:
1 - although the license seems perfectly fine, introducing another different license into json-c is an unnecessary burden for anyone that needs to analyze them before using the software
2 - to be complete, if we were to consider the changes you're contributing to be significant enough to warrant including a new license, it should include a copyright statement.

RyDroid:
I removed the license notice.

PR #254, opened 2016-08-08
Very minor changes to some tests

RyDroid:
- C89 comment > C99 comment
- Less long lines (for small screens and see 2 files as the same with a "multiple panels" feature like the one in GNU Emacs and Qt Creator)
- return STDC_CONST for main
- Errors to stderr
- Other things

### DIFF
--- a/tests/test-defs.sh
+++ b/tests/test-defs.sh
@@ -124,5 +124,3 @@ run_output_test()
 
 	return $err
 }
-
-
--- a/tests/test1.c
+++ b/tests/test1.c
@@ -78,14 +78,14 @@ void test_array_del_idx()
 		       rc, json_object_to_json_string(my_array));
 	}
 
-	// One more time, with the empty array:
+	/* One more time, with the empty array: */
 	rc = json_object_array_del_idx(my_array, 0, 1);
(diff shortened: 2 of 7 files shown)


hawicz:
This looks generally ok, but there are a couple of changes in there that look like they'll affect the output of the tests.  i.e. output to stderr in test4.c, and elimination of a newline in test_null.c.
Why are those needed?

RyDroid:
How could be a problem to print errors to stderr instead of stdout? It is the current behavior that seems strange to me.

puts had a new line. http://www.cplusplus.com/reference/cstdio/puts/
I fixed the place where I did the confusion with fputs.

hawicz:
Re: puts, ok.

As for stdout vs stderr, the tests write their output to a file and then that file is compared with the expected output.  Currently, only stdout is redirected, so if you start emitting messages to stderr it will no longer be possible to see what happened by simply looking at the output file.
That could be fixed by redirecting both stdout and stderr, but then what's the point of changing the output to go to stderr in the first place?

RyDroid:
I amended the commit in order not to change the stream of the prints.

PR #272, opened 2016-09-25
WIP: Allows for compilation with CMake on macOS

sgerbino:
…difies CMakeLists.txt to use them

This follows the method used by Win32 CMakeLists for macOS.


hawicz:
Down this path lies the madness of a whole pile of hand curated config.h files.  Now you add config.h.mac, then later it'll be config.h.mac.sierra, config.h.mac.panther, then config.h.mac.sierra_with_update_foo_installed, config.h.debian8.3_with_package_x, etc...
That's stupid when we have a perfectly functional way to figure these out automatically: the configure script.
If you want to change the cmakefile to run configure, then I'll include that, but I'm not merging this pr as-is.

sgerbino:
Makes sense... I'm going to chop away at it because I need this and I don't want to stray from your repository.

sgerbino:
I have it working now using autotools for UNIX systems and the existing method for Win32.
Everything seems to be in working order with Travis CI and Appveyor exception casting warnings with Windows x64. Is that supported?

PR #273, opened 2016-09-27
fixed wrong object name in json_object_all_values_equal

konoal:


hawicz:
d'oh!  Clearly this is need of some tests.  Thanks for the fix.

Issue #245, opened 2016-07-29
json-c compile in window xp


RafsanjaniRahman:
Hi, Just want to know if json-c lib can support compilation in window xp?

hawicz:
If you're asking whether itks possible to build json-c, I have no idea.  You'll likely need to make some adjustments, and if they're minimal enough I might accept them as pull requests.  I suspect this will largely depend on how you're trying to build (Visual Studio, cygwin gcc, mingw?).
However, insofar as any platform can be considered "supported", Windows XP is not on that list and you'll be largely on your own making it work.  I'll be happy to help if I can, but I certainly don't have an XP machine anywhere and I suspect there are few who do.

Nzbuu:
It certainly used to be possible to compile on Windows: https://ci.appveyor.com/project/Nzbuu/json-c/build/master.27

However, it now looks like this is broken. https://ci.appveyor.com/project/Nzbuu/json-c/build/master.29

hawicz:
Seems to be c++ stupidity not allowing void.  Do you know how to get that to build there as a C program?

jkaster:
This is an issue for both Win32 and x64 platforms. I tried limiting scope to Win32 but the problem still occurs.
```C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\stdio.h(346) : see declaration of '_snprintf'
json_object.c(723): error C2664: 'int json_object_double_to_json_string_format(json_object *,printbuf *,int,int,const char *)' : cannot convert argument 5 from 'void *' to 'const char *' [d:\dev\json-c\json-c.vcxproj]
Conversion from 'void*' to pointer to non-'void' requires an explicit cast```

I tried building from the command-line:
`msbuild json-c.vcxproj /p:configuration=debug /p:platform=Win32`
`msbuild json-c.vcxproj /p:configuration=release /p:platform=Win32`

jkaster:
It compiles if you change `json_object.c` lines 722:723 from:
```	return json_object_double_to_json_string_format(jso, pb, level, flags,
							jso->_userdata);```

to
```	return json_object_double_to_json_string_format(jso, pb, level, flags,
							(char *)jso->_userdata);```

hawicz:
It seems you're getting the "error" because you're compiling it as a c++ program, not a c one. 
[...]

PR #277, opened 2016-10-06
Json object set xxx

### LINKED ISSUE #269: json_object_<type>_set(json_object *o,<type> value) API for value setting in json object private structure
This will be really useful if you have a JSON representation of some internal state and when the state changes at the moment there is no way to change the JSON representation other than recreating the whole JSON tree which is suboptimal.
  Implementation is quite easy if you don't switch actual object types which will be enough for may specific itch that I'm willing to scratch it is OKed by the mainteners :smile:

sdrsdr:
extend the API to allow primitive object types to be set after construction.  this will also fix #269


sdrsdr:
looks like this will also fix #245

PR #282, opened 2016-11-01
ignore temporary files

cblp:
Temporary files are being regenerated often and polluting `git status`.


ploxiln:
worth noting that these files are not present in the master branch, and may be intentionally kept in the maintenance 0.12 branch - https://github.com/json-c/json-c/commit/ef2240098d73c0737d0a1bee6372fcf91756b7d4

cblp:
Yes, it may be something wrong with my build environment configuration.

ploxiln:
This is probably autotools "maintainer mode". If you get the source extracted from the tarball, the modification times of the files are such that configure has the same timestamp as configure.ac, or newer, etc for Makefile.in and Makefile.am. But if you checkout from git, the files are created in an arbitrary order, and on a filesystem with better-than second precision, it's likely that one of these files will be older than the file it is derived from. Then autoconf etc. will try to re-generate it.

One fix for this is to touch all files with a specific timestamp argument so they all have the same modification time. I think the `./configure --disable-dependency-tracking` option could also work.

hawicz:
Yes, those files *are* intentionally added on release branches.  Not merging this change.

PR #283, opened 2016-11-03
json_pointer: add first revision based on RFC 6901

commodo:
RFC link: https://tools.ietf.org/html/rfc6901


json_pointer.c
```diff
@@ -0,0 +1,242 @@
+/*
+ * $Id: json_pointer.c,v 1.20 2016/10/25 03:24:50 mclark Exp $
+ *
+ * Copyright (c) 2016 Metaparadigm Pte. Ltd.
```
hawicz: These copyrights are wrong.  If you wrote the code, your name should be on the copyright.

json_pointer.c
```diff
@@ -0,0 +1,242 @@
...
+}
+
+static int is_valid_index(struct json_object *jo, const char *path, int32_t *idx)
+{
+	int i, len = strlen(path);
+	/* this code-path optimizes a bit, for when we reference the 0-9 index range in a JSON array
+	   and because leading zeros not allowed */
+	if (len == 1) {
+		int ret = isdigit(path[0]);
+		ret ? (*idx = (path[0] - '0')) : (errno = EINVAL);
```
hawicz: What about the ENOENT case?

json_pointer.h
```diff
@@ -0,0 +1,59 @@
...
+#define _json_pointer_h_
+
+#include "json_object.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Retrieves and JSON sub-object from inside another JSON object
```
hawicz: s/and/a/

hawicz:
This looks pretty good.  Please address the four things I commented on in the changes, and I'll merge this.
* Fix copyrights
* ENOENT case in is_valid_index
* Slight typo in json_pointer.h, and 
* better explanation of ref counting behavior.

json_pointer.c
commodo: damn ; you're right ; thanks for catching this ; 
i think i added this after a quick refactor during writing this ; `is_valid_index()` initially was meant to handle if `path` is a valid number conforming to RFC ; but then i added the json_object param

json_pointer.h
commodo: good catch

commodo:
re-spinned ; with addressed comments; 
thanks :)

to give a bit o background [ if it helps ], we want to use this on top of OpenWrt/LEDE
json-c is used in a lot of other libs/modules in OpenWrt/LEDE
and we use OpenWrt/LEDE as base for our FW

the end-game is JSON Patch : https://tools.ietf.org/html/rfc6902
to get to JSON Patch, we need JSON Pointer first (which funny enough is RFC6901)

from the looks of it, the JSON Patch defines a JSON format for adding/removing/modifying JSON objects
and it also defines (briefly) the HTTP communication between server & client

now, for JSON pointer it looks reasonable to request to add it to libjson-c
[your opinion] what about JSON patch ? the non-HTTP part I mean ; 

depending on your input, I can go about it in 2 ways:
1. submit a PR with JSON Patch (the operations as defined in RFC6902) ; so, only the non-HTTP parts [ but this is after I implement JSON Patch ; I have nothing yet ]
2. create a new lib that uses libjson-c

i'm fine with either ; depending on your preference

thanks again
Alex

hawicz:
JSON Patch sounds like a reasonable addition to json-c.  A PR to add it would be most welcome.

PR #288, opened 2016-11-19
Fix _GNU_SOURCE define for vasprintf

andrenarchy:
Without the `#define` it errors with:
```
error: implicit declaration of function ‘vasprintf’ [-Werror=implicit-function-declaration]
     if((size = vasprintf(&t, msg, ap)) < 0) { va_end(ap); return -1; }
```

### DIFF
--- a/printbuf.c
+++ b/printbuf.c
@@ -13,6 +13,7 @@
  * (http://www.opensource.org/licenses/mit-license.php)
  */
 
+#define _GNU_SOURCE
 #include "config.h"
 
 #include <stdio.h>


andrenarchy:
I used cmake for building. It seems like this is not the preferred way of building this project. Feel free to close this PR if cmake is not yet supported.

hawicz:
Dropping that define in printbuf.c isn't the right way to fix this.  Building the normal way (configure, then make) results in -D_GNU_SOURCE being passed on the command line.  If you want to fix the cmake build, please try to figure out how to get it to use the same flags that the normal way of building does.

PR #289, opened 2016-11-27
bugfix: floating point representaion without fractional part 

### LINKED ISSUE #278: Serialization of double with no fractional component drops trailing zero
Calling `double-to-json_string` on a double with no fractional component results in the dropping of both the trailing zero and the period (e.g. `1.0` is serialized as `1`).   So when that string is deserialized, the type is inferred as an int.  I am not sure if this is the intended behavior of this function or not.  [This comment](https://github.com/json-c/json-c/blob/master/json_object.c#L716) leads me to believe it is not.  It seems that the use of `%g` is the cause of this behavior.

FWIW, you can find the full discussion we had about this issue in the flux-framework project [here](https://github.com/flux-framework/flux-sched/issues/214#issuecomment-256104514).

jgerhards:
closes #278


ploxiln:
It may be appropriate to move the which-format-string logic into a couple separate lines. Otherwise, looks good to me.

jgerhards:
I could change it if the maintainer wishes me to do so, however I don't deem it as necessary.

hawicz:
This seems like it might be a bit fragile.  I would think that checking the generated output for a '.' character would be more reliable, but if this works then let's go with it.

jgerhards:
as specified in man snprintf, this is guaranteed to work, see last sentence:
```
 g, G   The  double argument is converted in style f or e (or F or E for
              G conversions).  The precision specifies the number of  signifi‐
              cant  digits.   If the precision is missing, 6 digits are given;
              if the precision is zero, it is treated as 1.  Style e  is  used
              if  the  exponent from its conversion is less than -4 or greater
              than or equal to the precision.  Trailing zeros are removed from
              the  fractional part of the result; a decimal point appears only
              if it is followed by at least one digit.
```

PR #290, opened 2016-12-02
duplicate an json_object

chennqqi:
I add an function json_object_duplicate to duplicate an json_object. Do you accept this PR?


hawicz:
First of all, this doesn't compile due to the mismatched function prototype, but even if you fix that the approach of serializing the object tree into a string and then parsing it back out, while obviously convenient, is not one that I will accept into the library.
Instead, the code should walk the tree and copy each object directly.  The recently added json_c_visit() function may be a useful way to do this.

PR #294, opened 2016-12-12
Adding UTF-8 validation.  Fixes #122

### LINKED ISSUE #122: Add utf-8 validation when parsing strings.
php -r 'var_dump(json_decode("\"\xff\""));'
string(1) "�"

According to the maintainer Remi Collet, the fix for this needs to go in json-c:

https://github.com/remicollet/pecl-json-c/issues/9

Since json is used extensively these days for all kinds of services, this is a rather bad bug. Is there any way we can get this fixed in json-c, so the fix can move upstream to the distros using it?

vote539:
I modified the json_escape_str function to perform lightweight UTF-8 validation in addition to the string escaping that it already performs.  I added test cases for the new UTF-8 validation code.  Note that this code does not check for where the byte sequence is valid but the codepoint to which it maps to does not exist (the pink cells on [this Wikipedia page](https://en.wikipedia.org/wiki/UTF-8#Codepage_layout)).


vote539:
I don't know whether you want to make this change be the default or be something that you need to manually enable with a flag.  The fact that *json-c* can produce output containing invalid UTF-8 means that *json-c* can produce output that is invalid JSON, violating the contract of the library.  That being said, since this could potentially be a breaking change to users who depended on the old functionality, so it could go either way.

hawicz:
Thank you for the patch, but this doesn't seem to be adding any validation, instead it's munging the output of data that is already stored in memory within the json_object structure.  That's a rather different thing than what Issue #122 is talking about.  I would expect that if you're saying you're validating something, and in particular if you're claiming to be fixing Issue #122, then you'd be implementing changes to cause json_tokener_parse() to return an error of some sort.  To be entirely complete, json_object_new_string() would also fail in some way.
Instead, silently converting the invalid UTF-8 sequences to the UTF-8 replacement character (EF BF BD) when re-serializing the object is just going to make the behavior more difficult to understand since it'll neither match the input, nor the in-memory bytes.

btw, the use of "\0" within your string constants is pointless, there's already a terminating nul byte present.

vote539:
You're right, I misread the ticket #122 as being on the JSON generation step, not the JSON parsing step.

In any case, this patch addresses the related issue of sanitizing strings to valid UTF-8 upon serializing a JSON object.  Replacing invalid byte sequences with U+FFFD (EF BF BD) is standard behavior, and in fact it is what happens anyway when a forgiving JSON parser, like the default JSON parser in Node.JS, reads the output from json-c.  Stricter JSON parsers fail outright upon reading the invalid UTF-8.  Users who want to represent binary data in JSON should be using base64 or a similar encoding for their data and should not be expecting their arbitrary byte sequences to be saved in the serialized JSON strings.

Up to you if you want to accept the patch.  I have it applied to my local copy of json-c and it's been doing the job for me.

hawicz:
Thanks, but I think I'm going to pass on this patch.

PR #299, opened 2017-02-02
Improve json_object -> string performance

qlyoung:
Hi there! This is a performance-related patch to improve `json_object` -> string conversion.
It yields average performance gains of ~350% for large objects.

Breakdown
----------
A `perf` analysis of a tight loop around `json_object_to_json_string()` yields the following results:

flamegraph_bad

As you can see the primary hotspot here is `sprintbuf()` - more specifically, `vsnprintf()` / `vfprintf()`. Digging deeper, it appears that every call to `sprintbuf()` results in a full init of a variadic list. Variadic functions in themselves are pretty slow; they become rather onerous in terms of performance when called in a tight loop, as is often the case when dumping large amounts of JSON. I took a closer look at the rest of `json-c` and found that almost every call to `sprintbuf()` does not make use of the formatting it wraps -- indeed, most calls pass a single argument:

```
~/json-c> rg "sprintbuf"
json_object.c
150:                            sprintbuf(pb, "\\u00%c%c",
363:    sprintbuf(pb, "{" /*}*/);
365:            sprintbuf(pb, "\n");
370:                    sprintbuf(pb, ",");
372:                            sprintbuf(pb, "\n");
376:                    sprintbuf(pb, " ");
378:            sprintbuf(pb, "\"");
381:                    sprintbuf(pb, "\": ");
383:                    sprintbuf(pb, "\":");
385:                    sprintbuf(pb, "null");
393:                    sprintbuf(pb, "\n");
397:            return sprintbuf(pb, /*{*/ " }");
399:            return sprintbuf(pb, /*{*/ "}");
540:            return sprintbuf(pb, "true");
541:    return sprintbuf(pb, "false");
588:    return sprintbuf(pb, "%" PRId64, jso->o.c_int64);
856:    sprintbuf(pb, "\"");
858:    sprintbuf(pb, "\"");
975:    sprintbuf(pb, "[");
977:            sprintbuf(pb, "\n");
983:                    sprintbuf(pb, ",");
985:                            sprintbuf(pb, "\n");
989:                    sprintbuf(pb, " ");
993:                    sprintbuf(pb, "null");
1001:                   sprintbuf(pb, "\n");
1006:           return sprintbuf(pb, " ]");
1007:   return sprintbuf(pb, "]");
```

I've omitted `tests/` and `printbuf.[ch]` in this output.

Thus it's a lot of overhead to initialize a variadic list, attempt a write to a stack buffer, dynamically allocate a heap buffer if it isn't big enough etc. for every string write to a `printbuf`. This is especially so in the case of single character writes such as those for `{`, `}`, `[`, `]`.

Incidentally, this function appears to be an adaptation of a function described here: https://wavepackage.wordpress.com/tag/vsnprintf/

Which notes:

"*If you are not too concerned about performance*, this solution suits well gives you a comfortable way of manipulating strings." (sic)

Anyway my changes are pretty straightforward. Since the variadic part of `sprintbuf()` is only used in two places (besides tests), I removed that and just made `sprintbuf()` a wrapper around `printbuf_memappend()` that takes strlen for you. The two variadic prints have been replaced with `snprintf()`'s into static buffers that call `sprintbuf()` on the result.

This results in performance gains of about 350%, and much nicer flamegraph:

flamegraph_good

Some cursory `time` analysis backs this up.

Before:

```
~/json-c> time ./a.out
3.51user 0.00system 0:03.54elapsed 99%CPU (0avgtext+0avgdata 1920maxresident)k
0inputs+0outputs (0major+238minor)pagefaults 0swaps
~/json-c> time ./a.out
3.52user 0.00system 0:03.53elapsed 99%CPU (0avgtext+0avgdata 2012maxresident)k
0inputs+0outputs (0major+240minor)pagefaults 0swaps
~/json-c> time ./a.out
3.49user 0.00system 0:03.50elapsed 99%CPU (0avgtext+0avgdata 1932maxresident)k
0inputs+0outputs (0major+238minor)pagefaults 0swaps
```

After:

```
~/json-c> time ./a.out
0.99user 0.00system 0:01.00elapsed 99%CPU (0avgtext+0avgdata 1944maxresident)k
0inputs+0outputs (0major+237minor)pagefaults 0swaps
~/json-c> time ./a.out
1.02user 0.00system 0:01.03elapsed 99%CPU (0avgtext+0avgdata 1960maxresident)k
0inputs+0outputs (0major+240minor)pagefaults 0swaps
~/json-c> time ./a.out
1.02user 0.00system 0:01.03elapsed 99%CPU (0avgtext+0avgdata 1956maxresident)k
0inputs+0outputs (0major+238minor)pagefaults 0swaps
```

The test program I used for the flame graphs and these time results is:

```C
#include <stdio.h>
#include <json-c/json.h>

#define ITERATIONS 2

static json_object *
recursive_add_a2z (json_object *obj, unsigned int n)
{
  char key[5] = "KeyX";
  for (char i = 'A'; i < 'Z'; i++)
  {
    key[3] = i;
    if (n < ITERATIONS) {
      json_object *level = json_object_new_object();
      json_object_object_add (obj, key, recursive_add_a2z (level, n+1));
    }
    else {
      json_object *ex_string = json_object_new_string("Example");
      json_object_object_add(obj, "lol", ex_string);
      return obj;
    }
  }
  return obj;
}

int main (int argc, char *argv[])
{
  json_object *root = json_object_new_object();
  recursive_add_a2z (root, 0);
  for (int i = 0; i < 10000; i++)
  {
    json_object_to_json_string (root);
  }
}
```

For the flame graphs, the `for()` loop in main was changed to an infinite `while()` and allowed to run under `perf` for ~15 seconds.

I also removed the unit test for `sprintbuf()` since the tests for `printbuf_memappend()` now cover its functionality.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -147,9 +147,13 @@ static int json_escape_str(struct printbuf *pb, const char *str, int len, int fl
 					printbuf_memappend(pb,
 							   str + start_offset,
 							   pos - start_offset);
-				sprintbuf(pb, "\\u00%c%c",
-				json_hex_chars[c >> 4],
-				json_hex_chars[c & 0xf]);
+				static char sbuf[7];
+				snprintf(sbuf, sizeof(sbuf),
+                                        "\\u00%c%c",
+                                        json_hex_chars[c >> 4],
+                                        json_hex_chars[c & 0xf]);
+				printbuf_memappend (pb, sbuf, sizeof(sbuf) - 1);
+
 				start_offset = ++pos;
 			} else
 				pos++;
@@ -585,7 +589,10 @@ static int json_object_int_to_json_string(struct json_object* jso,
 					  int level,
 					  int flags)
 {
-	return sprintbuf(pb, "%" PRId64, jso->o.c_int64);
+	/* room for 19 digits, the sign char, and a null term */
+	static char sbuf[21];
+	snprintf(sbuf, sizeof(sbuf), "%"PRId64, jso->o.c_int64);
+	return sprintbuf(pb, sbuf);
 }
 
 struct json_object* json_object_new_int(int32_t i)
--- a/json_object.c
+++ b/json_object.c
@@ -149,11 +149,10 @@ static int json_escape_str(struct printbuf *pb, const char *str, int len, int fl
 							   pos - start_offset);
 				static char sbuf[7];
 				snprintf(sbuf, sizeof(sbuf),
-                                        "\\u00%c%c",
-                                        json_hex_chars[c >> 4],
-                                        json_hex_chars[c & 0xf]);
-				printbuf_memappend (pb, sbuf, sizeof(sbuf) - 1);
-
+					 "\\u00%c%c",
+					 json_hex_chars[c >> 4],
+					 json_hex_chars[c & 0xf]);
+				printbuf_memappend_fast(pb, sbuf, (int) sizeof(sbuf) - 1);
 				start_offset = ++pos;
 			} else
 				pos++;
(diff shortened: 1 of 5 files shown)


hawicz:
Wow, that's great, thanks for this PR!  However, there's at least one problem: this change will break backwards compatibility with code that's been built for earlier versions of the library.  (printbuf.h is part of the public API), so sorry, I can't accept the change as is.
However, instead of removing sprintbuf, how about adding a new macro:

```
#define _printbuf_ensure_string_constant(mystr) ("" mystr)
/**
 * ...define as operating only on string literals, 
 * and point reader at printbuf_memappend_fast w/ strlen, 
 * or sprintbuf for other uses...
 */
#define printbuf_strappend(pb, str) \
   printbuf_memappend_fast(pb, _printbuf_ensure_string_constant(str), (sizeof(str) - 1))

```
This means an extra 25 or so lines changed (not including tests), but I believe it's just a simple search and replace.

qlyoung:
Good suggestions. The macros also reduce function call overhead which adds a bit more performance :-)

Let me know how it looks now.

hawicz:
Merged.  The only question I have is your comment on sprintbuf, where you say:
"in these scenarios, consider using snprintf() with a static
+ * buffer in conjunction with one of the printbuf_*append() functions."

If you're in a situation where you need to use snprintf anyway, why avoid sprintfbuf?  Since snprintf is a variadic function too it seems like you're not going to get any performance benefit.

qlyoung:
Yeah, the comment was intended to be more about using a static buffer if the size is known and less about `sprintbuf()`. Looking again they should be roughly the same; did a little bit of testing and the static buffer is slightly faster but not really enough to make a sizable difference.

Thanks for your quick work on this PR!

PR #319, opened 2017-04-27
Windows: Fix dynamic library build with Visual Studio

dhenot:
This allows to compile json-c as a dynamic library with Visual Studio. This is needed because declaring a function as `extern` is not enough to make this compiler export it, so the resulting library was not exporting anything. I also export a couple functions which weren't before, as I was getting undefined references otherwise.

As far as I know this should not break anything on other platforms but I didn't have time to confirm that.

### DIFF
--- a/json_object.h
+++ b/json_object.h
@@ -21,6 +21,12 @@
 #define THIS_FUNCTION_IS_DEPRECATED(func) func
 #endif
 
+#if defined(_MSC_VER) 
+#define JSON_EXPORT __declspec(dllexport)
+#else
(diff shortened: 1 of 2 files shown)


HarmWibier:
It seems that this is preventing me from statically linking the json-c library with Visual Studio. 

I was successfully using the json-c library and I am trying to update due to other changes but this seems to prevent me from doing that.

HarmWibier:
I am sorry, I got it working now... I had some compiler settings wrong...

Issue #298, opened 2017-01-13
Building using CMake doesn't work


SanjoDeundiak:
I'm trying to build json-c with CMake (I want to add it as dependency to my project using CMake). I'm getting following error:
```

json-c/build$ make -j10

Scanning dependencies of target json-c
make[2]: Warning: File 'CMakeFiles/json-c.dir/depend.make' has modification time 2,4 s in the future
[ 20%] Building C object CMakeFiles/json-c.dir/arraylist.c.o
[ 20%] Building C object CMakeFiles/json-c.dir/debug.c.o
[ 30%] Building C object CMakeFiles/json-c.dir/json_object.c.o
[ 60%] Building C object CMakeFiles/json-c.dir/json_pointer.c.o
[ 60%] Building C object CMakeFiles/json-c.dir/json_tokener.c.o
[ 60%] Building C object CMakeFiles/json-c.dir/linkhash.c.o
[ 70%] Building C object CMakeFiles/json-c.dir/json_util.c.o
[ 80%] Building C object CMakeFiles/json-c.dir/printbuf.c.o
[ 90%] Building C object CMakeFiles/json-c.dir/random_seed.c.o
/media/sf_sources/sdmp-test/json-c/printbuf.c: In function ‘sprintbuf’:
/media/sf_sources/sdmp-test/json-c/printbuf.c:131:16: error: implicit declaration of function ‘vasprintf’ [-Werror=implicit-function-declaration]
     if((size = vasprintf(&t, msg, ap)) < 0) { va_end(ap); return -1; }
                ^
/media/sf_sources/sdmp-test/json-c/json_pointer.c: In function ‘json_pointer_getf’:
/media/sf_sources/sdmp-test/json-c/json_pointer.c:210:7: error: implicit declaration of function ‘vasprintf’ [-Werror=implicit-function-declaration]
  rc = vasprintf(&path_copy, path_fmt, args);
       ^
cc1: all warnings being treated as errors
CMakeFiles/json-c.dir/build.make:230: recipe for target 'CMakeFiles/json-c.dir/printbuf.c.o' failed
make[2]: *** [CMakeFiles/json-c.dir/printbuf.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
cc1: all warnings being treated as errors
CMakeFiles/json-c.dir/build.make:134: recipe for target 'CMakeFiles/json-c.dir/json_pointer.c.o' failed
[...]

PR #321, opened 2017-05-03
build,cmake: fix vasprintf implicit definition and generate both static & shared libs

commodo:
Fixes https://github.com/json-c/json-c/issues/298


commodo:
maybe it's a bit over-board/a lot
but i also squeezed into this PR the change for cmake to generate both static & shared lib
apparently, it was only generating static libs before

ploxiln:
Does this generate `libjson-c-static.a` instead of `libjson-c.a`? Meaning the name of the static library is different?

commodo:
yep ;
i admit i did not put much effort in trying to identify a way to see how to generate a static lib that's called `libjson-c.a`

will look into it

Issue #226, opened 2016-04-09
There is no json_object_new_null()


ilyash:
The only way I see to encode nulls is via `json_object_array_put_idx()` and it appears that nulls can not be encoded anywhere else (like `{"a": null}` for example). Please add support for encoding nulls uniformly, as any other type.

ilyash:
Looking at the source code of `json_object_to_json_string_ext()` I've found that one can just pass NULL to get the effect. Just could find it in any examples on the net. I guess if there was a `json_object_new_null()` it would help discoverability.

dota17:
> Issue #226 is basically asking to have the "null" type be treated the same as the other types, and implement it as an actual instance of a struct json_object, instead of a C NULL value.
There are many benefits to that consistency, but the problem with doing this is that it would be a major change to the API, as any existing code that was checking for null by comparing against NULL would no longer work the same.
The theoretically problematic json_type_null case isn't reachable since you can't actually allocate an object with type json_type_null (using the public API; json_object_new() is static).
The call to json_object_set_serializer() isn't actually valid to do on a NULL object (it'll abort at the assert in json_object_set_userdata()), which is just one example of what might be improved if we were to make null a first class json-c type.
The most we can do while maintaining API and ABI compatibility would be to, as @ilyash mentioned, create a json_object_new_null() "for visibility", but it would be just:

@hawicz As you said in PR  #428, for maintaining API and ABI compatibility, we had better to add a json_object_new_null() which just returns NULL. Or do nothing. I can submit this pr and close this issue. What's your opinion on this?

```
struct json_object *json_object_new_null()
{
    return NULL;
}
```

ilyash-b:
> as any existing code that was checking for null by comparing against NULL would no longer work the same.

I understand the need for compatibility. I recommend not to perpetuate the problem - it will continue to cost to developers and many users of the library. New major version maybe?

> implement it as an actual instance of a struct json_object

Yes please! That's the right thing to do, we all know that, right? :P

dota17:
json_object_new_null() has been added and this issue could be closed @hawicz

PR #328, opened 2017-06-16
json_object: implement json_object_deep_copy() function

commodo:
Seems to perform better than outputting to string and re-parsing it.

```
BENCHMARK - 1000000 iterations of 'dst2 = json_tokener_parse(json_object_get_string(src2))' took 20 seconds
BENCHMARK - 1000000 iterations of 'dst2 = json_tokener_parse(json_object_get_string(src2))' took 7 seconds
```

This function will be needed for JSON patch [RFC 6902] ; before applying a patch, a backup of the original object will be made ; in case any patch operation fails, the JSON object should not be modified at all.
This is covered in section `5.  Error Handling`

This PR also adds the `json_object_new_null()` function (which is requested here https://github.com/json-c/json-c/issues/226 ).
While copying the JSON object, it's easier to have an object-ual representation for the JSON `null`.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -1313,3 +1313,106 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
 	return 0;
 }
 
+static int json_object_copy_serializer_data(struct json_object *src, struct json_object *dst)
+{
+	/* FIXME: this is shared between copies ; maybe add a `_user_copy` cb here */
+	dst->_userdata = src->_userdata;
+	dst->_to_json_string = src->_to_json_string;
+	dst->_user_delete = src->_user_delete;
+	return 0;
+}
+
+static int json_object_deep_copy_recursive(struct json_object *src, struct json_object **dst)
+{
+	struct json_object *jso = NULL;
+	struct json_object_iter iter;
+	size_t i;
+
+	if (!src || !dst) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	switch (src->o_type) {
+		case json_type_boolean:
+			*dst = json_object_new_boolean(src->o.c_boolean);
+			break;
+
+		case json_type_double:
+			*dst = json_object_new_double(src->o.c_double);
+			break;
+
+		case json_type_int:
+			*dst = json_object_new_int64(src->o.c_int64);
+			break;
+
+		case json_type_string:
+			*dst = json_object_new_string(get_string_component(src));
+			break;
+
+		case json_type_object:
+			*dst = json_object_new_object();
+			if (!*dst) {
+				errno = ENOMEM;
+				return -1;
+			}
+			json_object_object_foreachC(src, iter) {
+				/* This handles the `json_type_null` case */
+				if (!iter.val)
+					jso = NULL;
+				else if (json_object_deep_copy_recursive(iter.val, &jso) < 0)
+					return -1;
+				if (json_object_object_add(*dst, iter.key, jso) < 0)
+					return -1;
+			}
+			break;
+
+		case json_type_array:
+			*dst = json_object_new_array();
+			if (!*dst) {
+				errno = ENOMEM;
+				return -1;
+			}
+			for (i = 0; i < json_object_array_length(src); i++) {
+				struct json_object *jso1 = json_object_array_get_idx(src, i);
+				/* This handles the `json_type_null` case */
+				if (!jso1)
+					jso = NULL;
+				else if (json_object_deep_copy_recursive(jso1, &jso) < 0)
+						return -1;
+				if (json_object_array_add(*dst, jso) < 0)
+					return -1;
+			}
+			break;
+
+		default:
+			errno = EINVAL;
+			return -1;
+	};
+
+	/* errno should be set by the function that faulted */
+	if (!*dst)
+		return -1;
+
+	return json_object_copy_serializer_data(src, *dst);
+}
+
+int json_object_deep_copy(struct json_object *src, struct json_object **dst)
+{
+	int rc;
+
+	/* Check if arguments are sane ; *dst must not point to a non-NULL object */
+	if (!src || !dst || *dst) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	rc = json_object_deep_copy_recursive(src, dst);
+	if (rc < 0) {
+		json_object_put(*dst);
+		*dst = NULL;
+	}
+
+	return rc;
+}
+
--- a/json_object.h
+++ b/json_object.h
@@ -979,6 +979,24 @@ JSON_EXPORT int json_object_set_string_len(json_object* obj, const char* new_val
 JSON_EXPORT int json_object_equal(struct json_object *obj1,
 			     struct json_object *obj2);
 
+/**
+ * Copy the contents of the JSON object.
+ * The destination object must be initialized to NULL,
+ * to make sure this function won't overwrite an existing JSON object.
+ *
+ * This does roughly the same thing as
+ * `json_tokener_parse(json_object_get_string(src))`.
+ *
+ * @param src source JSON object whose contents will be copied
+ * @param dst pointer to the destination object where the contents of `src`;
+ *            make sure this pointer is initialized to NULL
+ *
+ * @returns 0 if the copy went well, -1 if an error occured during copy
+ *          or if the destination pointer is non-NULL
+ */
+
+extern int json_object_deep_copy(struct json_object *src, struct json_object **dst);
+
 #ifdef __cplusplus
 }
 #endif
--- a/tests/test_deep_copy.test
+++ b/tests/test_deep_copy.test
@@ -1,0 +1,1 @@
+test_basic.test
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@
 /tests/test_cast
 /tests/test_charcase
 /tests/test_compare
+/tests/test_deep_copy
 /tests/test_double_serializer
 /tests/test_float
 /tests/test_json_pointer
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -13,6 +13,7 @@ TESTS+= test2.test
 TESTS+= test4.test
 TESTS+= testReplaceExisting.test
 TESTS+= test_parse_int64.test
+TESTS+= test_deep_copy.test
 TESTS+= test_null.test
 TESTS+= test_cast.test
 TESTS+= test_double_serializer.test
--- a/tests/test_deep_copy.c
+++ b/tests/test_deep_copy.c
@@ -1,0 +1,175 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stddef.h>
+#include <string.h>
+#include <assert.h>
+#include <errno.h>
+#include <time.h>
+
+#include "json.h"
+
+static const char *json_str1 = 
+"{"
+"    \"glossary\": {"
+"        \"title\": \"example glossary\","
+"        \"GlossDiv\": {"
+"            \"title\": \"S\","
+"            \"null_obj\": null, "
+"            \"GlossList\": {"
+"                \"GlossEntry\": {"
+"                    \"ID\": \"SGML\","
+"                    \"SortAs\": \"SGML\","
+"                    \"GlossTerm\": \"Standard Generalized Markup Language\","
+"                    \"Acronym\": \"SGML\","
+"                    \"Abbrev\": \"ISO 8879:1986\","
+"                    \"GlossDef\": {"
+"                        \"para\": \"A meta-markup language, used to create markup languages such as DocBook.\","
+"                        \"GlossSeeAlso\": [\"GML\", \"XML\"]"
+"                    },"
+"                    \"GlossSee\": \"markup\""
+"                }"
+"            }"
+"        }"
+"    }"
+"}";
+
+static const char *json_str2 =
+"{\"menu\": {"
+"    \"header\": \"SVG Viewer\","
+"    \"items\": ["
+"        {\"id\": \"Open\"},"
+"        {\"id\": \"OpenNew\", \"label\": \"Open New\"},"
+"        null,"
+"        {\"id\": \"ZoomIn\", \"label\": \"Zoom In\"},"
+"        {\"id\": \"ZoomOut\", \"label\": \"Zoom Out\"},"
+"        {\"id\": \"OriginalView\", \"label\": \"Original View\"},"
+"        null,"
+"        {\"id\": \"Quality\", \"another_null\": null},"
+"        {\"id\": \"Pause\"},"
+"        {\"id\": \"Mute\"},"
+"        null,"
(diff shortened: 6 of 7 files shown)


commodo:
I thought a bit about the `json_object_new_null()` change.
This might have a slight change, which should not be a problem with `json_object_is_type(jso, json_type_null)` and `json_object_is_type(json_object_new_null(), json_type_null)`  returning both `true`.

I don't feel this should be a problem in the whole JSON context.
After taking a quick look through the [JSON RFC](https://tools.ietf.org/html/rfc4627) I did not find anything to enforce the null-ness case.

But, if there's another opinion here, I'm open for discussion.

tests/test_null.c
```diff
@@ -31,7 +30,7 @@ int main(void)
 		printf("parsed string:   ");
 		puts(json);
 		puts("FAIL");
-		retval=1;
+		return 1;
```
hawicz: Why are you bailing early?  I expect that this will cause memory leak errors.

tests/test_null.c
```diff
@@ -52,6 +51,25 @@ int main(void)
...
+		json_object_put(jso);
+		json_object_put(null_object);
+		return 1;
+	}
+
+	json_object_object_add(jso, "a_null_object", null_object);
+	printf("%s\n", json_object_get_string(jso));
+
+	json_object_put(null_object);
+	puts("PASS");
```
hawicz: This should be left out, since whether it actually passed depends on whether all of the output matches what's in the .expected file.

hawicz: Or rather, we should adjust the "PASS" lines here to more accurately indicate that it's just the individual sections of the test that passed.

json_object.c
```diff
@@ -1214,3 +1214,98 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+		return -1;
+
+	return rc;
+}
+
+int json_object_deep_copy(struct json_object *src, struct json_object **dst)
+{
+	int rc;
+	/* Nothing to do, so return succes */
+	if (!src && !dst)
```
hawicz: IMO !dst should always be an error.  If you're not providing some place to put the copied object you're doing something wrong.

json_object.c
```diff
@@ -1214,3 +1214,98 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+		return 0;
+
+	/* Check if arguments are sane ; refcount must not be higher than 1 */
+	if (!src || !dst || (*dst && (*dst)->_ref_count > 1)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	/* Free previous destination object */
+	json_object_put(*dst);
```
hawicz: skip this if *dst == NULL

json_object.c
```diff
@@ -1214,3 +1214,98 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+
+int json_object_deep_copy(struct json_object *src, struct json_object **dst)
+{
+	int rc;
+	/* Nothing to do, so return succes */
+	if (!src && !dst)
+		return 0;
+
+	/* Check if arguments are sane ; refcount must not be higher than 1 */
+	if (!src || !dst || (*dst && (*dst)->_ref_count > 1)) {
```
hawicz: If we're going to allow *dst to be set at all, then it seems unnecessarily limiting to have the ref count restriction here.  I'd say just json_object_put it if there's something there, do a *dst = NULL, and continue on.

hawicz:
Having 'null' in the JSON that you parse turn into something other than a C NULL is a fairly significant API change.  If we were to really switch to having 'null' values show up as json_type_null C objects then we need to go whole-hog and start treating NULL as an invalid value in many places (such as json_object_object_add), and have a major library version bump.
The fact that 'null' values show up as C NULL's will make the deepcopy code a bit more complicated, but it doesn't seem like it should be all that bad, and trying to tie implementing deepcopy to all the API changes needed to really support json_type_null as an object seems like it's asking for trouble.

json_object.h
```diff
@@ -899,6 +899,28 @@ extern int json_object_set_string_len(json_object* obj, const char* new_value, i
 extern int json_object_equal(struct json_object *obj1,
 			     struct json_object *obj2);
 
+/**
+ * Copy the contents of the JSON object.
+ * If the destination object is NULL, it will be allocated.
+ * If the destination object is non-NULL, it will be overwrriten
+ * completely, if the refcount is 1.
+ *
+ * Note: the same result could be achieved by using
```
hawicz: This could be reworded a bit.  Perhaps just:

> This does roughly the same thing as `json_tokener_parse(json_object_get_string(src))`

json_object.h
```diff
@@ -899,6 +899,28 @@ extern int json_object_set_string_len(json_object* obj, const char* new_value, i
 extern int json_object_equal(struct json_object *obj1,
 			     struct json_object *obj2);
 
+/**
+ * Copy the contents of the JSON object.
+ * If the destination object is NULL, it will be allocated.
+ * If the destination object is non-NULL, it will be overwrriten
```
hawicz: How about:
> If *dst is non-NULL, it will be passed to json_object_put (which may free it if the refcount drops to 0)
> and a completely new object will be created and assigned to *dst.

json_object.c
```diff
@@ -1214,3 +1214,98 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+		default:
+			errno = EINVAL;
+			rc = -1;
+			break;
+	};
+
+	/* errno should be set by the function that faulted */
+	if (!*dst)
+		return -1;
+
```
hawicz: This needs to copy over at least the serializer and data used by json_object_new_double_s(), and ideally any user-specified ones.  However, for user-specified ones we have the problem that we can free userdata, but have no way to copy it. hmm...

tests/test_null.c
commodo: the idea here was that if the test fails here, stop it here, so that we don't generate more output
but, since the output is compared against an expected output the test should fail;
will fix

json_object.c
commodo: ack

json_object.c
commodo: but, if `refcount  > 1`, then it implies that the object may be shared by some other entity;
overwriting it in that case may cause weird/unexpected behavior for any other entity holding a refcount to this object

we can drop the refcount check [if you really insist] ;
i'd feel a bit uncomfortable about it, though, but i could live with it

json_object.c
commodo: `json_object_put()` handles null;
adding a null-check here may yield a small performance improvement, in the sense that a jump to the `json_object_put()` block, won't happen if that is `null`

most modern compilers would probably optimize this call 
not sure about older compilers & deployment [of json-c on them]

hmm ; i guess i'm neutral about adding or not-adding it;
instead of over-thinking it, i'll just add it

json_object.h
commodo: ack

json_object.h
commodo: before addressing this, let's resolve this comment first :)
https://github.com/json-c/json-c/pull/328/commits/9f420cdf94b1c83800cae66b22f035f8182ac5c3#r126346219

json_object.c
commodo: ack
i had no idea about that serializer data;

commodo:
> Having 'null' in the JSON that you parse turn into something other than a C NULL is a fairly significant API change. If we were to really switch to having 'null' values show up as json_type_null C objects then we need to go whole-hog and start treating NULL as an invalid value in many places (such as json_object_object_add), and have a major library version bump.
> The fact that 'null' values show up as C NULL's will make the deepcopy code a bit more complicated, but it doesn't seem like it should be all that bad, and trying to tie implementing deepcopy to all the API changes needed to really support json_type_null as an object seems like it's asking for trouble.

i think i could implement a deep-copy that may work without this null JSON object change ; 
i am no longer as sure [how possible it is] as i would have been when i first implemented the change tho ;
but I could try to re-visit this ;

the null object change, is mostly a personal ambition that i tried to do ;
i agree the API change is significant ; 
maybe i could re-implement the deep copy without this, and re-visit the whole null object thing after a version bump ;

commodo:
@hawicz 
* dropped `json_object_new_null()` function ; maybe will re-visit this later [ let's see ]
* added copying for serializer data [from json_object_new_double_s() ] ; it's function `json_object_copy_serializer_data()` ; please check if you agree with it it :)
* add null check to `*dst`
* returning EINVAL now ; even if both !src  & !dst

Still need to resolve the discussion [and related comment] if  `(*dst)->_refcount > 1`.

json_object.c
```diff
@@ -1265,3 +1265,128 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
 	return 0;
 }
 
+static int json_object_copy_serializer_data(struct json_object *src, struct json_object *dst)
+{
+	/* `src` and `dst` were validated by now ==> no null check for them */
+	if (src->o_type != json_type_double)
+		return 0;
+	/* Nothing to do */
+	if (!src->_userdata)
```
hawicz: _userdata doesn't actually need to be non-NULL.  You can set a serializer that doesn't need userdata.

json_object.c
```diff
@@ -1265,3 +1265,128 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+		return 0;
+	if (!src->_user_delete) {
```
hawicz: Please drop this check too.  As with _userdata, _user_delete is optional.

json_object.c
```diff
@@ -1265,3 +1265,128 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
 	return 0;
 }
 
+static int json_object_copy_serializer_data(struct json_object *src, struct json_object *dst)
+{
+	/* `src` and `dst` were validated by now ==> no null check for them */
+	if (src->o_type != json_type_double)
```
hawicz: Although we currently only use json_object_userdata_to_json_string for double's, there's no particular reason it couldn't be used for other types as well, and if it's in place we really don't care what type the object is, so please drop this check.

json_object.c
```diff
@@ -1265,3 +1265,128 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+			break;
+
+		case json_type_object:
+			*dst = json_object_new_object();
+			if (!*dst)
+				return -1;
+			json_object_object_foreachC(src, iter) {
+				/* This handles the `json_type_null` case */
+				if (!iter.val) {
+					json_object_object_add(*dst, iter.key, NULL);
```
hawicz: How about have this line be "jso = NULL;", drop the continue, and change 1331 to "else if ...".  That way we have a single call to json_object_object_add, and the return value is checked.

json_object.c
```diff
@@ -1265,3 +1265,128 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+					continue;
+				}
+				if ((rc = json_object_deep_copy_recursive(iter.val, &jso)) < 0)
```
hawicz: Did you intend to do something with rc before returning?  If not, don't bother using it.  It looks like the only place it's actually used at all is right at the end of this function, and even there it's not really needed.

json_object.c
```diff
@@ -1265,3 +1265,128 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
...
+
+		case json_type_array:
+			*dst = json_object_new_array();
+			if (!*dst)
+				return -1;
+			for (i = 0; i < json_object_array_length(src); i++) {
+				struct json_object *jso1 = json_object_array_get_idx(src, i);
+				/* This handles the `json_type_null` case */
+				if (!jso1) {
+					json_object_array_add(*dst, NULL);
```
hawicz: Same comment here as for the json_type_object case.

tests/test_deep_copy.test
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
hawicz: Please change this to be a symlink to the newly added test_basic.test

hawicz:
Did you respond about the (*dst)->_refcount > 1 check somewhere?  Github isn't showing that to me (nor my original comment :( )
Anyway, I think that check should be dropped.  Either it should support the normal reference count handling, where passing something with _refcount==1 frees the object, and _refcount>1 just decrements by 1, or it should enforce that *dst is NULL, with my preference being the former.

tests/test_deep_copy.test
commodo: ack ;
missed it through all the rebasing

commodo:
>Did you respond about the (*dst)->_refcount > 1 check somewhere? Github isn't showing that to me (nor my original comment :( )

It was this comment series: 
* your comment https://github.com/json-c/json-c/pull/328#discussion_r126289096
* my reply  https://github.com/json-c/json-c/pull/328#discussion_r126346219

Github is kinda crappy after a long review ; Bitbucket is s bit better, but not by much.
On Github, per-line comments get imploded after a force-push, because the lines are outdated [so you need to manually find them and expand them ].
One way to go about it, is to close a PR and open a new one [and reference the old PR/comments] ; after a few force pushes, it gets even harder to find old comments.
[ semi-offtopic: i see a business opportunity to build a per-line-of-codes review tool that is better/easier to use ]

> Anyway, I think that check should be dropped. Either it should support the normal reference count handling, where passing something with _refcount==1 frees the object, and _refcount>1 just decrements by 1, or it should enforce that *dst is NULL, with my preference being the former.

Regarding `Enforcing that *dst is NULL, with my preference being the former.`
I can agree with this.

For the `_refcount == 1` enforcement, I had in mind the ability to do some sort of construct like:
```
struct json_object dst = NULL;
for ( ;; ) {
   < logic1 >
   json_deep_copy(src, &dst)
   < logic2 >
}
```
So, in this case you can rewrite the `*dst` object.
The thought came to be during the JSON Patch work. A JSON Patch `replace` operation would simply be a `json_deep_copy(src, &dst)`.

But I can understand [what you're saying] that, as a standalone function, `json_deep_copy()` would better enforce that `*dst` is NULL, and let the caller handle free-ing/prep-ing the `*dst` object.
[ Assuming, that's the direction you're aiming for ]

json_object.c
commodo: ack

json_object.c
commodo: ack;
not sure anymore ; 
the `rc` is mostly out of habit, when I try to use more `goto` statements ; which is not the case
will remove `rc` altogether

json_object.c
commodo: ack

json_object.c
commodo: so, regarding other/external serializers, maybe a `_user_copy` function would be needed

commodo:
what if we change
```
extern int json_object_deep_copy(struct json_object *src, struct json_object **dst); 
```
to
```
extern struct json_object *json_object_deep_copy(struct json_object *src); 
```
?

commodo:
re-spinned with review comments addressed
[ hopefully all of them [that were made so far]
feel free to add more

PR #331, opened 2017-06-25
Safe use of snprintf() / vsnprintf() for Visual studio, and thread-safety fix

rouault:
Two unrelated issues but touching the same are of code:
- For Visual Studio, we use the '#define snprint _snprintf' macro, but
 _snprinft isn't C99 compliant, and in particularly will not NUL terminate
 if the buffer isn't large enough. So 0 terminate explicitly
- In json_escape_str / json_object_int_to_json_string a static char[] buffer is
  used. This isn't thread-safe. Just use a plain char[]

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -774,17 +774,16 @@ static int json_object_double_to_json_string_format(struct json_object* jso,
 		size = snprintf(buf, sizeof(buf), format, jso->o.c_double);
 		buf[sizeof(buf)-1] = '\0';
 
-		if (modf(jso->o.c_double, &dummy) == 0)
+		if (modf(jso->o.c_double, &dummy) == 0 && size >= 0 && size < (int)sizeof(buf) - 2 )
 		{
 			// Ensure it looks like a float, even if snprintf didn't.
-			strncat(buf, ".0", sizeof(buf) - 1);
-			if (size >= 0)
-				size += 2; // yes, even if strncat ran out of room
+			strcat(buf, ".0");
+			size += 2;
 		}
 	}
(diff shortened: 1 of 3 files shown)


json_object.c
```diff
@@ -589,8 +590,9 @@ static int json_object_int_to_json_string(struct json_object* jso,
 					  int flags)
 {
 	/* room for 19 digits, the sign char, and a null term */
-	static char sbuf[21];
+	char sbuf[21];
```
qlyoung: ouch, MT didn't even cross my mind here... thanks for cleaning up after me

hawicz:
ew... if snprintf and vsnprintf are broken on some platform we shouldn't hack around it in the code, but instead provide fixed versions (or wrappers).  i.e. #define snprintf json_c_snprintf and write a json_c_snprintf that does the right thing.

rouault:
> #define snprintf json_c_snprintf and write a json_c_snprintf that does the right thing.

That can be pretty complicated. C99 compliant snprintf() should return the theroretical length of the string that should be provided to avoid truncation... MSVC _snprintf() just returns -1 in that case.

hawicz:
At the very least you can add the null termination there, instead of repeating it all over the rest of the code and risking missing a snprintf location, or having someone add a new call to snprintf.

hawicz:
I think these changes need to be untangled a bit.  I'd like to merge over the fix for the bad strncat usage and the thread safety fix now, but can't.  I'll probably just re-do those individually.

hawicz:
Equivalent changes have been done in individual commits, without merging this, so I'm closing this.

PR #333, opened 2017-07-03
Replace obsolete AM_PROG_LIBTOOL

besser82:
This macro is obsolete since at least autoconf-2.63.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -87,7 +87,7 @@ int main(int c,char* v) {return 0;}
 
 AC_LANG_POP([C])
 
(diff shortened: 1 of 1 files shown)


qlyoung:
@hawicz this looks good to me, cf. libtool docs:

https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html

hawicz:
Given that we're now saying that you *need* to have a recent libtool (and autoconf, and automake) installed, this is ok to pull in now.

PR #336, opened 2017-07-10
tests: fix tests in travis-ci.org

commodo:
Seems some failing tests have piled up, due to a missing `set -e`

* `test_util_file` test ; the issue is that Apple linkers don't allow to override `strerror` and this seems to be locked down on travis-ci.org [ for security reasons I assume ] ; I tried some solutions from the web (like here: http://tlrobinson.net/blog/2007/12/overriding-library-functions-in-mac-os-x-the-easy-way-dyld_insert_libraries/) ; patches can be checked out here: https://github.com/commodo/json-c/commit/49ed91eb39d96ac625553843ba2c02018f601964  & https://github.com/commodo/json-c/commit/ca13901e0c37a9885fbc665c638185f0ef1c804f  ; so, trying to force the test's `strerror`  did not work for me ; they worked locally, but not on travis ;  I gave up [ after a few hours ] and resumed to go by checking specific output only for Darwin
* `test_locale` : not sure if the fix is sane ; seems that `setlocale()` actually changes the locale on Apple, while the `uselocale()` function does not ; 

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,6 +23,22 @@ else
   AC_MSG_RESULT([RDRAND Hardware RNG Hash Seed disabled. Use --enable-rdrand to enable])
 fi
 
+AC_ARG_ENABLE(strerror-override,
+ AS_HELP_STRING([--enable-strerror-override],
+   [Override strerror() function with internal version.]),
+[if test x$enableval = xyes; then
+  enable_strerror_override=yes
+  AC_DEFINE(ENABLE_STRERROR_OVERRIDE, 1, [Override strerror() with internal version])
+fi])
+
+AM_CONDITIONAL([ENABLE_STRERROR_OVERRIDE], [test "x$enable_strerror_override" = "xyes"])
+
+if test "x$enable_strerror_override" = "xyes"; then
+  AC_MSG_RESULT([Overriding `strerror()` function with internal version])
+else
+  AC_MSG_RESULT([Using libc's `strerror()` function])
+fi
+
 # enable silent build by default
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
--- a/configure.ac
+++ b/configure.ac
@@ -7,6 +7,8 @@ AM_INIT_AUTOMAKE
 
 AC_PROG_MAKE_SET
 
+AC_CANONICAL_HOST
+
 AC_ARG_ENABLE(rdrand,
  AS_HELP_STRING([--enable-rdrand],
    [Enable RDRAND Hardware RNG Hash Seed generation on supported x86/x64 platforms.]),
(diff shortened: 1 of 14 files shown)


commodo:
re-spinned to latest master;

json_tokener.c
```diff
@@ -239,7 +239,7 @@ struct json_object* json_tokener_parse_ex(struct json_tokener *tok,
 {
   struct json_object *obj = NULL;
   char c = '\1';
-#ifdef HAVE_USELOCALE
+#if defined(HAVE_USELOCALE) && !defined(__APPLE__)
```
hawicz: If there's some problem with uselocale or duplocale then the configure test should be adjusted to check for that instead.

tests/test_util_file.expected.Darwin
```diff
@@ -0,0 +1,35 @@
+OK: json_object_from_fd(./valid.json)={ "foo": 123 }
```
hawicz: Bleh, that sucks about not being able to replace strerror(), but rather than ending up with per-arch output files, let's instead s/strerror/_json_c_strerror/ and then have the tests replace that.

json_tokener.c
commodo: ack

tests/test_util_file.expected.Darwin
commodo: ack;
this will have to be done via a preprocessor `#define` at build time in `.travis.yml` ; or maybe via a special `configure` rule ;
because `libjson-c` is built first and that would link to `strerror`

i also thought about renaming `strerror`
wasn't sure what your thoughts/input would be

commodo:
@hawicz 
re-spinned 
for the `strerror` override, i chose the way of the "build option"

feel free to disagree with/comment on any of my approaches :)
i'm not completely sure that i chose the best approach[es] ;
well, maybe it's more of a preference thing ;
in which case, i prefer that your preference takes precedence ;

configure.ac
```diff
@@ -57,14 +75,22 @@ AS_IF([test "x$ac_cv___thread" != xno],
...
-AC_CHECK_FUNCS(strcasecmp strdup strerror snprintf vsnprintf vasprintf open vsyslog strncasecmp setlocale uselocale)
+AC_CHECK_FUNCS(strcasecmp strdup strerror snprintf vsnprintf vasprintf open vsyslog strncasecmp setlocale)
 AC_CHECK_DECLS([INFINITY], [], [], [[#include <math.h>]])
 AC_CHECK_DECLS([nan], [], [], [[#include <math.h>]])
 AC_CHECK_DECLS([isnan], [], [], [[#include <math.h>]])
 AC_CHECK_DECLS([isinf], [], [], [[#include <math.h>]])
 AC_CHECK_DECLS([_isnan], [], [], [[#include <float.h>]])
 AC_CHECK_DECLS([_finite], [], [], [[#include <float.h>]])
 
+case "${host_os}" in
```
hawicz: Why should linux be the only OS allowed to use uselocale?  When I said that the configure test needs to be adjusted I mean that it needs to actually check whether uselocale works.  This is effectively just an obfuscated #ifdef __APPLE__.
I'm going to merge this PR since all the other changes look fine, but then I'm going to revert this piece and we'll need to approach it a different way.

hawicz:
ack, wait a sec, this break the tests ALL the time.  :(
Having that work conditional on a special config option isn't going to work.

commodo:
when you said about the configure option [instead of the #ifdef APPLE  define ] I assumed you have something against `#ifdef __PLATOFRM_DEFINES__` ;
the `#ifdef APPLE` seemed to shortest path to the fix 

so, yeah, the `uselocale` function seems to behave differently on OS X than Linux [ at least in the `test_locale` test ]
for the `setlocale` I don't know what's different [if anything] between OS X & Linux

now, i'm also wondering whether it would have worked to fix the test [ add check #ifdef HAVE_USELOCALE #elsif HAVE_SETLOCALE  [ same as in in libjson-c] ]
or, maybe that would have just hidden the issue ;

we can insist a bit more on this `locale` issue
to be honest, i'm just glad tests are back on track ; for me, the locale thing is a bit secondary atm ;
because i would like to get back to JSON patch [ first stop : `json_deep_copy()` ]
and i'm also thinking about checking into JSON schema later [will see]

there is one element that may crop up later [from the LEDE/OpenWrt/Embedded side of things] ;
some of the stuff that's being added now, grows the lib a bit ;
so, maybe i'll propose some  --enable/--disable json-patch/pointer/schema/etc   ; 
but let's see about this when we get there ; it may not be needed ;

hawicz:
The configure test you added effectively *is* platform define, rather than a feature check.  Whether you use #ifdef APPLE or #ifdef linux or ${host_os} = "linux" you have the same problem: it's only loosely related to the behavior that you're actually trying to adapt to

I'm working on fixing the tests now, since they are definitely not on track with these changes.  If we can use the function override functionality, then _json_c_strerror just needs to always be used, with a flag to choose which output you get.

PR #355, opened 2017-08-22
1.make it can been compiled with Visual Studio 2010 by modify the CMakeList.txt and others

Haffon:
2.replace json_object_get/put API with json_object_retain/release, as they operate the reference counter, and confused with array_list_get/put_idx.
3.replace array_list_get/put_idx API with array_list_get/insert to make them more clear to use.
4.removed the lh_get_hash static decoration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,8 @@ set(JSON_C_HEADERS
     ./linkhash.h
     ./math_compat.h
     ./strdup_compat.h
+	./strerror_override.h
+	./strerror_override_private
     ./vasprintf_compat.h
     ./printbuf.h
     ./random_seed.h
@@ -66,6 +68,7 @@ set(JSON_C_SOURCES
     ./json_util.c
     ./linkhash.c
     ./printbuf.c
+	./strerror_override.c
     ./random_seed.c
 )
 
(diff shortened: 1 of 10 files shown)


hawicz:
Thanks for the pull request, but this is a severely API-breaking change, so I can't merge it, sorry.
Also, given that there are much more recent versions of VS available for free, I'm reluctant to make changes just to cater to such an old version.  However, if you want to re-spin a PR that only has simple changes, like moving where variables are declared, I'd be willing to merge it.

Haffon:
My English is not well, you can't merge this pull request is ok, let's say something more:
1,Both array_list_get/put_idx and json_object_get/put use the same suffix get/put, but their meaning is different.
2.I just think json-c never released a stable version, so changing API should be acceptable? Make API pretty until too later.
3.I prefer VS 2005-2010 just because their remote debugger no need install in target machine, it help me a lot to debug problem which only occur in one man's computer. I don't know why MS make it disabled in modern VS.

hawicz:
Although official releases of json-c have been rare, the API on the default branch IS stable and changing it is not acceptable.
Sorry, can't help with remote debugger setup (although it seems like in newer VS it's just a separate download).

Haffon:
@hawicz I re-spin the pull request, you can review it now.

PR #360, opened 2017-09-07
Use strtoll() to parse ints

jasonbking:
Instead of using `sscanf()` (whose behavior is surprisingly underspecified) to parse integer values, use `strtoll()` instead.  It turns out to be much simpler, and hopefully is more portable.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -173,6 +173,7 @@ AX_COMPILE_CHECK_SIZEOF(int)
 AX_COMPILE_CHECK_SIZEOF(long)
 AX_COMPILE_CHECK_SIZEOF(long long)
 AX_COMPILE_CHECK_SIZEOF(size_t, [#include <stdint.h>])
+AX_COMPILE_CHECK_SIZEOF(int64_t, [#include <stdint.h>])
 
 AC_CONFIG_FILES([
 Makefile
--- a/json_object.c
+++ b/json_object.c
@@ -33,6 +33,10 @@
 #include "strdup_compat.h"
 #include "snprintf_compat.h"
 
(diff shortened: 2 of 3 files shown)


jasonbking:
It's not obvious what the AppVeyor failures were -- nothing I pushed should have touched any VS project files.  Is there something more specific someone can point me at?

hawicz:
So how do we know sizeof(long long) == sizeof(int64_t) ?

(the AppVeyor build it probably because I removed the VS project files, and didn't do anything to make it use cmake instead)

jasonbking:
https://en.wikipedia.org/wiki/C_data_types has the (likely) more digestible version, but includes a link to the relevant standards.  tl;dr -- C99 requires that `long long` can hold at least a 64-bit signed integer.  No implementation I'm aware of today uses any integer size larger than that for `long long` (it'd likely break a lot of stuff), though if there are concerns of that happening, I could have it do something like:

```C
char *end = NULL;
long long val;
errno = 0;
val = stroll(buf, &end, 10);
if (end != buf) {
    if (val > INT64_MAX)
        *retval = INT64_MAX;
   else if (val < INT64_MIN)
       *retval = INT64_MIN;
   else
        *retval = val;
}
return ((val == 0 && errno != 0) || (end == buf)) ? 1 : 0;
```

hawicz:
You're right, it's pretty unlikely.  Enough so that I'd say instead of code to handle it let's just do a preprocessor check like:

    #if SIZEOF_LONG_LONG != SIZEOF_INT64_T
    #error The long long type isn't 64-bits
    #endif

PR #362, opened 2017-09-13
CMake Package Config

jwinarske:
These changes enable package config.


hawicz:
The changes to configure.ac should not be merged.
I think the creation of the .pc file using the existing json-c.pc.in file instead of adding another cmake specific one.
I won't be merging this, but I'll be checking in some equivalent changes in a bit.

hawicz:
See commit cf91362 for the pc file changes.

Issue #338, opened 2017-07-14
implement json_object_int_inc(json_object *, int64_t)


dpintaric:
Hi,

Having json_object_set_int is great since I don't have to json_object_object_del() first but if I just want to inc the value I have to:

`json_object_set_int(key_jobj, json_object_get_int(key_jobj) + 1);`

Having json_object_inc_int() and  json_object_inc_int64() would greatly optimize this when a interger object is beeing used as a counter.

hawicz:
Sure, sounds good to me.  A pull request would be welcome.

PR #363, opened 2017-09-14
Issue #338, add json_object_add_int functions

jvijtiuk:
This adds json_object_add_int and json_object_add_int64 functions, instead of the increment functions requested in #338 since add functions are more general.


json_object.h
```diff
@@ -718,6 +718,22 @@ JSON_EXPORT int32_t json_object_get_int(const struct json_object *obj);
...
+ * If the addition would result in a overflow, the object value
+ * is set to INT32_MAX.
+ * If the addition would result in a underflow, the object value
+ * is set to INT32_MIN.
+ *
+ * @param obj the json_object instance
+ * @param val the value to add
+ * @returns 1 if the addition succeded, 0 otherwise
+ */
+JSON_EXPORT int json_object_add_int(struct json_object *obj, int val);
```
commodo: maybe more of a preference [from me] here, but, keeping the name `json_object_inc_int()` [like mentioned in issue #338 ] or similar, could be a good idea, because it makes it clearer that it's incrementing an int value;

in the context of `libjson-c` a name that like `json_object_add_xxx` reminds me of the `json_object_object_add()` or `json_object_array_add()` functions where a JSON object is added to a another JSON object

i'd also rename the `int val` param to `int inc`
and you can also create a convenience macro called `json_object_dec_int()` where
```
#define json_object_dec_int(x,y)    json_object_inc_int(x, -(y))
```

but these are all just preferences/thoughts :)

hawicz:
Merging as-is first, but I'm going to consolidate the 32/64-bit versions shortly.

hawicz:
Also, I like @commodo suggestion of keeping the "inc" in the name, so I'll rename it.

hawicz:
Actually, to be consistent with other methods that mutate existing objects (e.g. json_object_array_sort) I'm going to use json_object_int_inc() here.

hawicz:
Done in commit 3628f16

PR #383, opened 2017-12-04
Fix a compiler warning.

ssrlive:
In VS 2015, the warning text is "warning C4550: expression evaluates to a function which is missing an argument list".


ssrlive:
BTW. I have find out lots of other definitions. for example "json_object_to_json_string_fn" etc.

PR #384, opened 2017-12-04
Fix a VS 2015 compiler warning.

ssrlive:
In VS 2015, the warning text is "warning C4550: expression evaluates to a function which is missing an argument list".


hawicz:
It looks like there are a few more places that could have the same problem.  I'll fix those too.

hawicz:
See commit aedd36a

ssrlive:
Thanks.

PR #388, opened 2017-12-11
json_object: Add size_t json_object_sizeof()

besser82:
This comes in handy when one wants to pre-allocate memory to store a bunch of `struct json_object`, like [gdal does](https://github.com/OSGeo/gdal/blob/f8c279b1c1ae1da1b5025e2a43a1b4d99acf2852/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp#L45).

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -509,6 +509,11 @@ int json_object_object_length(const struct json_object *jso)
 	return lh_table_length(jso->o.c_object);
 }
 
+size_t json_object_sizeof(void)
+{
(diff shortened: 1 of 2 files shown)


besser82:
@hawicz Can you please backport this to the 0.13 branch, so we can have this function in a possible future 0.13.1 release?

besser82:
@hawicz I opened #390 to backport this for 0.13.X series…

PR #389, opened 2017-12-13
json_object: Avoid double free (and thus a segfault) when ref_count gets < 0

besser82:
Without this patch `libu2f-server` and `sway` may segfault during regular code execution.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -182,6 +182,11 @@ int json_object_put(struct json_object *jso)
 {
 	if(!jso) return 0;
 
+	/* Avoid invalid free and crash explicitly instead of (silently)
+	 * segfaulting.
+	 */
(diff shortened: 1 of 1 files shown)


besser82:
Without this patch `libu2f-server` and `sway` may segfault during regular code execution.

hawicz:
er... if refcount gets below zero I think you *want* segfault.  Otherwise it's going to be pretty tough to figure out where something went wrong.

hawicz:
Rather than trying to work around the segfault due to double free, I think the better change would be to add a:

    assert(jso->_ref_count > 0);

near the beginning of json_object_put, so at least the crash is explicit.

besser82:
@hawicz I've made the change and rebased onto recent master.

hawicz:
Great, thanks!

PR #415, opened 2018-05-21
Resolve windows name conflict

sgbihu:
1. The windows dll will output the lib and dll, and rename the static
    lib will have conflict on windows.
2. Delete rename code to dismiss the conflict.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,9 @@ add_library(json-c-static
 
 set_property(TARGET json-c PROPERTY C_STANDARD 99)
 set_property(TARGET json-c-static PROPERTY C_STANDARD 99)
+if (NOT MSVC)
 set_target_properties(json-c-static PROPERTIES OUTPUT_NAME json-c)
+endif()
 
 install(TARGETS json-c json-c-static
     RUNTIME DESTINATION bin


hawicz:
How is it a conflict if one is called json-c.lib and the other json-c.dll?

hawicz:
Though I hate to close this without really understanding what you think the problem is, this doesn't seem to be a useful change to make, so I'm closing this without merging it.

sgbihu:
I'm sorry to reply you so late.

In shared library VS project, it generate two file for user, one is json-c.lib, another is json-c.dll. The lib is for link, and dll for execute.

So the project will build both static and shared library. The static configuration will output json-c.lib and that lib will be overwritten by shared configuration. Eventually, it will only generate shared build. 

Thanks.

hawicz:
I poked around a bit, and found a few things that explain a bit what the problem with windows linking is, e.g.:
https://blogs.msdn.microsoft.com/oldnewthing/20091013-00/?p=16403
https://stackoverflow.com/questions/6421693/why-are-lib-files-beasts-of-such-a-duplicitous-nature

Basically, windows leaves the information you need to actually link against a dll out of it, but puts it in a .lib file that is called an "import library".  Then, it re-uses the same .lib extension for that import library and for static libraries, so if you want your static and dynamic libraries to have the same name you're SOL.  Argh!

Anyway, it looks like the change you propose is a necessary and reasonable fix.

PR #428, opened 2018-07-07
Added new_null() function

abinpaul1:
There is no json_object_new_null() #226

I added a functon in json_object.c to create object null. Please check and tell me if it requires changes. Please excuse any mistakes since this is my first major contribution in github.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -1030,6 +1030,12 @@ struct json_object* json_object_new_string(const char *s)
 	return jso;
 }
 
+struct json_object* json_object_new_null()
+{
+         const char *s=null;
+         return json_object_new_string(s);
+}
+
 struct json_object* json_object_new_string_len(const char *s, int len)
 {
 	char *dstbuf;


json_object.c
```diff
@@ -1030,6 +1030,12 @@ struct json_object* json_object_new_string(const char *s)
 	return jso;
 }
 
+struct json_object* json_object_new_null()
+{
+         const char *s=null;
+         return json_object_new_string(s);
```
ploxiln: This `null` is typically written `NULL` in C.

When `json_object_new_string()` gets to the line `jso->o.c_string.len = strlen(s);` it will crash if `s` is `NULL`.

You don't need an allocated object for json `null`, you can just use C `NULL` pointer value for that.

abinpaul1: Thanks for the feedback. Can you please clarify as to what the issue 226 is. I was trying to work on that.

ploxiln: I don't fully understand #226. It says that you can call `json_object_to_json_string_ext(NULL, 0);` which returns the string "null".

Most places I look handle `NULL` (`(!jso1)` is equivalent to `(jso == NULL)`):

```c
int json_object_is_type(const struct json_object *jso, enum json_type type)
{
	if (!jso)
		return (type == json_type_null);
	return (jso->o_type == type);
}

enum json_type json_object_get_type(const struct json_object *jso)
{
	if (!jso)
		return json_type_null;
	return jso->o_type;
}
```

If you want an allocated json object of type null, you can do: `json_object_new(json_type_null)` - but this doesn't work out when re-serializing. You need to set the `_to_json_string` member function pointer to something appropriate, but I think it'll need a new one. Also, `json_object_set_serializer()` is wrong, it does:

```c
		// Reset to the standard serialization function
		switch(jso->o_type)
		{
		case json_type_null:
			jso->_to_json_string = NULL;
```

But the `...to_json_string()` functions do:

```c
	if (!jso)
	{
		s = 4;
		r = "null";
	}
	else if ((jso->_pb) || (jso->_pb = printbuf_new()))
	{
		printbuf_reset(jso->_pb);
		if(jso->_to_json_string(jso, jso->_pb, 0, flags) >= 0)
```

So you see, you _can_ allocate an object with type `json_type_null`, but it won't work right. However, a `NULL` pointer does seem to work as a json null most (maybe all) places.

hawicz: Issue #226 is basically asking to have the "null" type be treated the same as the other types, and implement it as an actual instance of a struct json_object, instead of a C NULL value.
There are many benefits to that consistency, but the problem with doing this is that it would be a major change to the API, as any existing code that was checking for null by comparing against NULL would no longer work the same.

The theoretically problematic json_type_null case isn't reachable since you *can't* actually allocate an object with type json_type_null (using the public API; json_object_new() is static).
The call to json_object_set_serializer() isn't actually valid to do on a NULL object (it'll abort at the assert in json_object_set_userdata()), which is just one example of what might be improved if we were to make null a first class json-c type.

The most we can do while maintaining API and ABI compatibility would be to, as @ilyash mentioned, create a json_object_new_null() "for visibility", but it would be just:

    struct json_object *json_object_new_null()
    {
        return NULL;
    }

(plus an appropriately explanatory doc comment, of course)

hawicz:
A trivial json_object_new_null() that simply returns NULL was added in commit 010f33d and we're not going to be adding an implementation that returns an actual object anytime soon if at all, so I'm closing this.

PR #432, opened 2018-07-22
The real CMake support.

unmanned-player:
This patch provides a CMakeLists.txt file to build JSON-C library without relying on auto-tools support making the build a bit more portable and cleaner.

It can detect headers and symbols and generate config.h header file based on those detections. It cannot yet handle ctest(1) as the testing itself depends on comparing the output against files. Testing would need some creative abuse of CMake :)

Hopefully, cross-compiling would get a lot easier.

The patch has been tested on GCC 4.8.5 (Linux), GCC 7.2.0 (MinGW) and Microsoft Visual C++ 16.0 (2010?) using CMake 2.8.12 and CMake 3.5.

PR #434, opened 2018-07-23
The real CMake support

unmanned-player:
This patch provides a CMakeLists.txt file to build JSON-C library without relying on auto-tools support. This makes the build a bit more portable and cleaner.

It can detect headers and symbols and generate config.h header file based on those detections. It cannot yet handle ctest(1) as the testing itself depends on comparing the output against files. Testing would need some creative abuse of CMake :) This will be provided a few patches later and may possibly involve re-factoring test cases.

The patch has been tested on GCC 4.8.5 (Linux), GCC 7.2.0 (MinGW) and Microsoft Visual C++ 16.0 (2010?) locally. Also, compiles correctly on Travis CI and AppVeyor without errors.

PS: Finally, this updates the patch provided in #432 & #433 by correcting AppVeyor file. Sorry for the inconvenience. I learnt what AppVeyor is and corrected the build automation too. There's a lot more to AppVeyor than what I initially thought.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,147 +1,253 @@
-#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+# Many projects still are stuck using CMake 2.8 is several places so it's good to provide backward support too. This is
+# specially true in old embedded systems (OpenWRT and friends) where CMake isn't necessarily upgraded.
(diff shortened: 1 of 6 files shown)


unmanned-player:
Did you have a chance to look at it? It is near equivalent to the one in master branch now.

Issue #429, opened 2018-07-08
Conflict of interest between JSON_C_TO_STRING_SPACED and JSON_C_TO_STRING_PRETTY


lhunath:
The goal of JSON_C_TO_STRING_SPACED is to include minimal spacing in the output.

The goal of JSON_C_TO_STRING_PRETTY is to indent the output.

JSON_C_TO_STRING_SPACED's minimal spacing includes spacing inside objects and spacing after colons.

JSON_C_TO_STRING_PRETTY's indenting includes spacing inside objects and newlines.

If you want output which has both minimal spacing and is indented (eg. indenting + space after colon), you need to enable both flags, but this currently causes bad indenting.

Therefore, I propose that if JSON_C_TO_STRING_PRETTY is enabled, JSON_C_TO_STRING_SPACED's object-spacing is ignored (because pretty already provides its own object spacing).  Alternatively, JSON_C_TO_STRING_PRETTY could be changed to be a complete subset of JSON_C_TO_STRING_SPACED's minimal spacing, which means it would also have to add a space after the colon.

As implemented right now, there is a strange conflict where pretty is not spacing + indenting, and to get all spacing features, your indenting cannot be done right.

lhunath:
```
diff --git json_object.c json_object.c
index 8a86bc6..c5bca7c 100644
--- json_object.c
+++ json_object.c
@@ -395,7 +395,7 @@ static int json_object_object_to_json_string(struct json_object* jso,
                                printbuf_strappend(pb, "\n");
                }
                had_children = 1;
-               if (flags & JSON_C_TO_STRING_SPACED)
+               if (flags & JSON_C_TO_STRING_SPACED && !(flags & JSON_C_TO_STRING_PRETTY))
                        printbuf_strappend(pb, " ");
                indent(pb, level+1, flags);
                printbuf_strappend(pb, "\"");
@@ -416,7 +416,7 @@ static int json_object_object_to_json_string(struct json_object* jso,
                        printbuf_strappend(pb, "\n");
                indent(pb,level,flags);
        }
-       if (flags & JSON_C_TO_STRING_SPACED)
+       if (flags & JSON_C_TO_STRING_SPACED && !(flags & JSON_C_TO_STRING_PRETTY))
                return printbuf_strappend(pb, /*{*/ " }");
        else
                return printbuf_strappend(pb, /*{*/ "}");
@@ -1127,7 +1127,7 @@ static int json_object_array_to_json_string(struct json_object* jso,
[...]

PR #453, opened 2018-11-20
Fixed misalignment in JSON string due to space after \n being printed...

darjankrijan:
...when choosing JSON_C_TO_STRING_SPACED together with JSON_C_TO_STRING_PRETTY in json_object_array_to_json_string

Previous code resulted in a misaligned "}" at the end and (1 + level\*2)spaces for flags&JSON_C_TO_STRING_PRETTY or 1 space + level\*tabs for flags&(JSON_C_TO_STRING_PRETTY|JSON_C_TO_STRING_PRETTY_TAB) being used for indentation.


hawicz:
This looks incomplete.  See issue #429 

Also, this really needs a test to verify that it's working as expected.

darjankrijan:
I'll have a closer look tonight.

darjankrijan:
I added the workaround for the array string function as well, also a test for the new changes.

I now actually wonder how test2Formatted_pretty passes. The .expected result seems to be perfectly aligned with level*2 spaces and no space before the last "}".

Anyway, I hope this fixes the issue.

tests/test_string_spaced_pretty.c
```diff
@@ -0,0 +1,34 @@
+#include <stdio.h>
```
hawicz: There's no need to create a whole new set of code for this.  test1Formatted and test2Formatted will both happily take multiple flags, though you'll need to tweak test2.test to do something like

    run_output_test -o test2Formatted_spaced_pretty test2Formatted spaced pretty

since the existing loop won't handle that.

hawicz:
test2Formatted_pretty passed because it's not trying to specify multiple formatting flags.  You only run into problems if you try to do *both* spaced and pretty.

darjankrijan:
I incorporated it into both test1 and test2.

hawicz:
Thanks for updating that!

PR #464, opened 2018-12-11
Speed up parsing and object creation

ramiropolla:
Around 40% speedup for 32-bit x86 and around 20% speedup for 64-bit x86.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -41,7 +41,6 @@
 // Don't define this.  It's not thread-safe.
 /* #define REFCOUNT_DEBUG 1 */
 
-const char *json_number_chars = "0123456789.+-eE";
 const char *json_hex_chars = "0123456789abcdefABCDEF";
 
 static struct json_object* json_object_new(enum json_type o_type);
--- a/json_object.c
+++ b/json_object.c
@@ -1388,7 +1388,7 @@ int json_c_shallow_copy_default(json_object *src, json_object *parent, const cha
 		break;
 
 	case json_type_string:
-		*dst = json_object_new_string(get_string_component(src));
+		*dst = json_object_new_string_len(get_string_component(src), src->o.c_string.len);
 		break;
 
 	case json_type_object:
--- a/json_object.c
+++ b/json_object.c
@@ -240,11 +240,14 @@ static struct json_object* json_object_new(enum json_type o_type)
 {
 	struct json_object *jso;
 
-	jso = (struct json_object*)calloc(sizeof(struct json_object), 1);
+	jso = (struct json_object*)malloc(sizeof(struct json_object));
 	if (!jso)
 		return NULL;
 	jso->o_type = o_type;
 	jso->_ref_count = 1;
+	jso->_pb = NULL;
+	jso->_user_delete = NULL;
+	jso->_userdata = NULL;
 #ifdef REFCOUNT_DEBUG
 	lh_table_insert(json_object_table, jso, jso);
 	MC_DEBUG("json_object_new_%s: %p\n", json_type_to_name(jso->o_type), jso);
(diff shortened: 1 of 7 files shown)


hawicz:
This PR has some valuable performance improvements, but there are a number of things that need to be fixed first before I can merge it:
* You need to declare functions first before defining them, even inline ones.  See the errors over at Appveyor.
* Add comments explaining why it's ok to switch away from the ctype functions (i.e. we're assuming ascii and only "C" locale)
* json_number_chars  should be removed, since with your changes we're no longer using it.
* Lots of problems with the int64 parsing:
  * Having two different ways to parse an int64 is error prone.  Please find a way to share the code, and ensure that the tests in test_parse_int64 exercise your code.  If inlining the code really makes that much of a performance difference (does it?) then stuff the code into a header and include it twice so we're at least using the same logic if not the exact same code.
  * parse_int64() isn't a complete 64-bit integer parsing function, and will give completely incorrect results if you pass it a string with non-digits present.  That could be ok, since I think how it's called in json_tokener.c guarantees that, but the function needs a comment clearly documenting that limitation.  On the other hand, other places where json_parse_int64() is used *don't* guarantee that.
  * The overflow handling in parse_int64() doesn't work.  For example, try parsing -9223372036854775809 or 9223372036854775808 or 88888888888888888888.  Even when it does cap it to the max or min value, it improperly returns success (0).
  * parse_int64() uses the wrong constants LLONG_MAX/LLONG_MIN, use INT64_MAX/INT64_MIN instead.

ramiropolla:
Thank you for the review.

I force pushed the changes to the branch on the pull request. I also made a branch with the fixes to the previous pull request (before merging the commits) so that they're easier to review <https://github.com/ramiropolla/json-c/tree/for_upstream-fixes>.

> This PR has some valuable performance improvements, but there are a number of things that need to be fixed first before I can merge it:
> 
> * You need to declare functions first before defining them, even inline ones.  See the errors over at Appveyor.

It seems the issue wasn't that the functions were missing forward declarations, but that older versions of Visual Studio don't support the "inline" keyword. Any compiler should end up inlining these small static functions, regardless of the inline keyword, so I removed the keyword.

> * Add comments explaining why it's ok to switch away from the ctype functions (i.e. we're assuming ascii and only "C" locale)

Done.

> * json_number_chars  should be removed, since with your changes we're no longer using it.

Done.

> * Lots of problems with the int64 parsing:
>   
>   * Having two different ways to parse an int64 is error prone.  Please find a way to share the code, and ensure that the tests in test_parse_int64 exercise your code.  If inlining the code really makes that much of a performance difference (does it?) then stuff the code into a header and include it twice so we're at least using the same logic if not the exact same code.

There was a tiny slowdown by using a non-inlined function, so I moved it to json_util.c and made json_parse_int64() use it. All tests still pass.

>   * parse_int64() isn't a complete 64-bit integer parsing function, and will give completely incorrect results if you pass it a string with non-digits present.  That could be ok, since I think how it's called in json_tokener.c guarantees that, but the function needs a comment clearly documenting that limitation.  On the other hand, other places where json_parse_int64() is used _don't_ guarantee that.

I added a comment.

>   * The overflow handling in parse_int64() doesn't work.  For example, try parsing -9223372036854775809 or 9223372036854775808 or 88888888888888888888.  Even when it does cap it to the max or min value, it improperly returns success (0).

The current behaviour is to return success on overflow (it did seem weird to me), so I left it as-is for now. This could be fixed in a separate PR. I fixed the checks for the corner cases though.

>   * parse_int64() uses the wrong constants LLONG_MAX/LLONG_MIN, use INT64_MAX/INT64_MIN instead.

Fixed.

hawicz:
It looks like I merged most of the changes from this quite some time ago, and then totally forgot that I had done so.  The current code includes:

- Drop use of the _delete function pointer
- Replace calloc w/ malloc+explicit field initialization
- Additional use of json_object_new_string_len()
- Eliminate use of ctype.h
 
But it does _not_ include the json_parse_sanitized_int64() and related changes.  It may be worth re-doing that as a separate PR.

PR #471, opened 2018-12-23
create folders with mode 0755 when building

a3nm:
Hi,

My user has umask `077`. When running `sudo make install`, the directories '/usr/local/include/json-c' and '/usr/local/lib/pkgconfig' are created with the corresponding permissions, which means that users of the system cannot read them afterwards.

This patch adds a setting to ensure that folders are created with the right permissions, inspired by this SO answer https://unix.stackexchange.com/a/436000. I have tested that it fixes the problem for me, i.e., the Makefile is generated with the corresponding mkdir calls, and running `sudo make install` creates the directories with the proper permissions.

(Disclaimer: I don't know anything about autotools, so I don't know if this is the right way to fix the problem (or right place to add the directive I added). Also I don't know whether this impacts portability, i.e., are there reasonable systems where `mkdir` does not support the `-m` flag?)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,6 +11,8 @@ AC_PROG_MAKE_SET
 
 AC_CANONICAL_HOST
 
+MKDIR_P="mkdir -m 0755 -p"
+
 AC_ARG_ENABLE(threading,
  AS_HELP_STRING([--enable-threading],
    [Enable code to support partly multi-threaded use]),


besser82:
> My user has umask `077`. When running `sudo make install`, the directories '/usr/local/include/json-c' and '/usr/local/lib/pkgconfig' are created with the corresponding permissions, which means that users of the system cannot read them afterwards.

Then you have an unusual configuration of your system.  Doing something different then declared by POSIX specification implies knowledge of its implications.

Doing some kludge patching in Autotools is *NOT* the right way.

I'd rather strongly recommend "fixing" such customizations of umask in another custom way.  In this case you should use: `make mkdir_p="mkdir -p -m 0755 --" install` to install applications that are build with Autotools.

besser82:
Well, I'm not the maintainer of this project nor a collaborator.

Anyways, this behauviour of Automake has been discussed some years ago and was discarded by its developers:  https://debbugs.gnu.org/cgi/bugreport.cgi?bug=12130

a3nm:
Interesting, thanks for the pointer.

Looking closely at that discussion, it seems like the discussion was closed in the last message in answer to the suggestion of improving `mkdir -p` (as a way to workaround the problem), but the question about autotools wasn't addressed in that discussion.

So yeah, I guess that this discussion should be resurrected with autotools... As for this PR, I'm leaving it open depending on whether the json-c maintainer wants to integrate my proposed (and admittedly hacky) workaround.

hawicz:
The change to configure.ac is rendered obsolete by the switch to cmake, but I made the equivalent cmake-based change in commit fa6bc1e.

a3nm:
Cool, thanks!

PR #476, opened 2019-02-21
Add new function named json_object_new_string_noalloc

makiftasova:
On some low memory systems (routers in my case), calling malloc (by the use of strdup) causes
memory fragmentation which eventually results in out of memory
situation. Thus eliminating this malloc call whenever possible improves the situation.

I know providing such "magic" API is a risk itself, thus I tried to explain how it works at the function doc in  json_object.h file.

I am also aware that implementation introduces code duplication but I couldn't find a better way, thus I am open for comments and suggestions.


hawicz:
This seems reasonable, but instead of duplicating the code, how about having a json_object_new_string_ext(char *s, size_t len, int flags) function, and use it for json_object_new_string(), json_object_new_string_len() and this new function?

makiftasova:
This one is a RFC commit, I didn't had the time to investigate why test_deep_copy test fails.

So if you can provide feedback about if I understand your de-duplication request correctly, it would be much appreciated.

besser82:
@makiftasova I think this PR can be closed, as I've opened a new PR, that iplements a working version of your code.

hawicz:
Sounds good to me, let's continue over at PR #620

PR #483, opened 2019-03-23
compatible with rarely-used Chinese characters in GBK charset

pj-x86:
compatible with rarely-used Chinese characters in GBK charset, those characters's the second byte is just the backslash '\'(0x5C). For example: 昞 (GBK encoding: 0x955C).
also see an example from json.txt
[json.txt](https://github.com/json-c/json-c/files/2999513/json.txt)

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -107,6 +107,8 @@ static int json_escape_str(struct printbuf *pb, const char *str, int len, int fl
 {
 	int pos = 0, start_offset = 0;
 	unsigned char c;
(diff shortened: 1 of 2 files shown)


ploxiln:
GBK charset is not necessarily valid json. The spec says "UNICODE character except `"` or `\`" and for those use escape sequences, among which is `\uXXXX` (4 hexadecimal digits) for any unicode codepoint.

pj-x86:
OK, I've got it. Thank you!

PR #484, opened 2019-05-05
Add support for uint64

emersion:
The JSON spec doesn't bound JSON numbers. In order to be able to store values
greater than INT64_MAX, adding uint64 support is required.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -177,7 +177,7 @@ extern struct json_object* json_object_get(struct json_object *jso)
 	__sync_add_and_fetch(&jso->_ref_count, 1);
 #else
 	++jso->_ref_count;
-#endif        
+#endif
 
 	return jso;
 }
@@ -193,7 +193,7 @@ int json_object_put(struct json_object *jso)
 
 #if defined(HAVE_ATOMIC_BUILTINS) && defined(ENABLE_THREADING)
 	/* Note: this only allow the refcount to remain correct
-	 * when multiple threads are adjusting it.  It is still an error 
+	 * when multiple threads are adjusting it.  It is still an error
 	 * for a thread to decrement the refcount if it doesn't "own" it,
 	 * as that can result in the thread that loses the race to 0
 	 * operating on an already-freed object.
@@ -481,7 +481,7 @@ int json_object_object_add_ex(struct json_object* jso,
 	// We lookup the entry and replace the value, rather than just deleting
 	// and re-adding it, so the existing key remains valid.
 	hash = lh_get_hash(jso->o.c_object, (const void *)key);
-	existing_entry = (opts & JSON_C_OBJECT_ADD_KEY_IS_NEW) ? NULL : 
+	existing_entry = (opts & JSON_C_OBJECT_ADD_KEY_IS_NEW) ? NULL :
 			      lh_table_lookup_entry_w_hash(jso->o.c_object,
 							   (const void *)key, hash);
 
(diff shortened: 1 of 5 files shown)


hawicz:
hmm... this:
* doesn't build in all cases (see AppVeyor)
* adds a whole extra field taking up space in the json_object structure
* litters the API with several extra methods
* doesn't include changes to do anything reasonable when retrieving a value using existing methods (i.e. everywhere else that json_type_int is referenced)
* doesn't actually parse larger integers in json_tokener
* includes unrelated whitespace changes in the same commit

Thanks for the patch, but I think a different approach is needed.

emersion:
>doesn't build in all cases (see AppVeyor)

Yup. This PR is a RFC.

>adds a whole extra field taking up space in the json_object structure that isn't used in existing code

That's not how union works. I don't know how you want json-c to use the extended field in existing code. Do you prefer to cast uint32_t to int32_t?

>litters the API with several extra methods

Well, int64 also "litters" the API.

>doesn't include changes to do anything reasonable when retrieving a value using existing methods (i.e. everywhere else that json_type_int is referenced)

Probably needs a flag in `json_object`?

>doesn't actually parse larger integers in json_tokener

Yup.

>includes unrelated whitespace changes in the same commit

Sigh. Should I submit these as a separate PR?

>Thanks for the patch, but I think a different approach is needed.

Suggestions welcome.

hawicz:
If you're going to add another field, instead of having it be a flag I think it would end up being a bit more natural for it to be a sign bit.  But if you're going to add a sign bit, you might as well use the other bits.  That implies adding support for 128-bit signed integers, and having uint64_t be a capped version of that.

So, the things I think would be needed for that are:
* Change c_int64 into a c_int64high and c_int64low (or a single 128-bit field, if that can be done portably enough)
* Adjust existing code that works with c_int64 to use the new fields
** Cap values similar to how json_object_get_int does, adjust max value constants used, etc...
* Adjust json_tokener to parse integers wider than 64-bits.
** Include some way to limit the allowable size of input integers
* Adjust json_object_int_to_json_string to emit wider integers
* Add _{new,get,set}_int128 and _{new,get,set}_uint64 methods
* Add tests to ensure the new code is working properly
* Benchmark, figure out whether the performance impact is enough to warrant making this a compile time option
* Mull over whether it's worth adding "just" 128-bit support, or whether it makes sense to go whole-hog and use a bignum library for arbitrary size integer support.

And yes, please split off whitespace changes to a separate commit (and ideally an independent pull request).

emersion:
>But if you're going to add a sign bit, you might as well use the other bits.

To be honest, I'm not a fan of having a non-symmetrical API wrt. signed/unsigned integers. There's no real reason to do so expect "json_type_int is signed".

What would you think of adding a `json_type_uint` type instead?

>So, the things I think would be needed for that are:

Yeah, that sounds like a good plan if we're going to implement it this way.

>And yes, please split off whitespace changes to a separate commit (and ideally an independent pull request).

Ack, will do.

ploxiln:
Another idea is to just note that int64 overflowed (e.g. a flag bit), and enable the user to access the original numeric string so they can feed it to their preferred bignum library (or `strtoull()` if they expect numbers bigger than INT64_MAX but smaller than UINT64_MAX).

besser82:
@hawicz I think this one can be closed, since we have support for `uint64` in `json-c` now.

emersion:
Superseded by https://github.com/json-c/json-c/pull/542

PR #485, opened 2019-05-21
Install CMake module files

myd7349:
This patch exports the CMake install target so that the users can install the CMake module files and use them in a CMake-based project like this:
```cmake
find_package(json-c CONFIG REQUIRED)
add_executable(json_c_test json_c_test.c)
target_link_libraries(json_c_test PRIVATE json-c::json-c)
```

The `json-c` target is put under `json-c::` namespace:
https://github.com/myd7349/json-c/blob/c2036ab9fc602cca8c9f200d6fe25d443b3b169e/CMakeLists.txt#L261

If:
```cmake
target_link_libraries(json_c_test PRIVATE json-c)
```
is more preferable, we may just delete that line.


myd7349:
https://github.com/microsoft/vcpkg/pull/6446

PR #490, opened 2019-06-07
do not compile with -Werror by default

### LINKED ISSUE #489: Don't use -Werror by default
Whilst I have no problem with the developer using `-Werror` to ensure that all compiler warnings are solved, please don't force this on everyone who wants to build json-c.

All it takes is one slightly changed dependency, or a newer compiler, or a different architecture, for there to be new warnings.  Latest example: json-c doesn't build in icecc because icecc strips out comments, so the switch fallthrough warnings are emitted, which are fatal because of `-Werror`.

ploxiln:
but do test build with `-Werror`

also fix `-Wwrite-string` to `-Wwrite-strings`

fixes #489

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_script:
   - ./configure
 
 script:
-  - make
+  - make AM_CFLAGS="-Werror"
 
 after_success:
   - make check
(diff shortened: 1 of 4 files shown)


ploxiln:
(oh, I guess that configure issue is what `-Wno-error=deprecated-declarations` was for ... honestly I didn't test with upstream master first, my bad)

ploxiln:
closing in favor of #491

PR #491, opened 2019-06-09
build: add option --disable-werror to configure

### LINKED ISSUE #489: Don't use -Werror by default

ploxiln:
a likely more welcome change than #490 based on comments in #489 

closes #489

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -165,7 +165,12 @@ AS_IF([test "x$enable_Bsymbolic" = "xcheck"],
 AS_IF([test "x$enable_Bsymbolic" = "xyes"], [JSON_BSYMBOLIC_LDFLAGS=-Wl[,]-Bsymbolic-functions])
 AC_SUBST(JSON_BSYMBOLIC_LDFLAGS)
 
-AX_APPEND_COMPILE_FLAGS([-Wall -Werror -Wcast-qual -Wno-error=deprecated-declarations])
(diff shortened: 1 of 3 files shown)


ploxiln:
a funny side-effect of `-Werror` is that without `-Werror`, the compile would produce these warnings:

```
warning: unknown warning option '-Wwrite-string'; did you mean '-Wwrite-strings'? [-Wunknown-warning-option]
```

but with `-Werror` added first, the configure check would decline to add `-Wwrite-string`:

```
checking whether C compiler accepts -Werror... yes
checking whether C compiler accepts -Wall... yes
checking whether C compiler accepts -Wcast-qual... yes
checking whether C compiler accepts -Wno-error=deprecated-declarations... yes
checking whether C compiler accepts -Wextra... yes
checking whether C compiler accepts -Wwrite-string... no
```

so there would be no indication that `-Wwrite-string` was wrong.

Another note - I can't do `(void*)"test"` unless I also add `-Wno-cast-qual`

PR #495, opened 2019-07-24
README.md:fix 2 typos

andy5995:


hawicz:
There's only one typo.  The phrase "...has some code to help make ***use in*** threaded programs a bit safer." is the intended meaning.  The code in json-c can't make a threaded program safer in general, but it can make use of json-c in a threaded program (somewhat) safer.

andy5995:
Understood. I reverted that change, but I wonder if **make its use in** or **make json-c's use in** would be a little more clearer? To my eyes, **make use in** doesn't flow quite right.

hawicz:
Sure, either of those ways sounds fine

PR #497, opened 2019-07-24
add common header for all tests

### LINKED ISSUE #406: Undefine NDEBUG for tests
Our build system defines `-DNDEBUG` via `CPPFLAGS` for release builds. Because of this, the tests don't compile. 

```
  CC       test2.o                                                                                                                                                                                        [21/1905]
  CCLD     test2
  CC       test4.o
  CCLD     test4
  CC       testReplaceExisting.o
  CCLD     testReplaceExisting
  CC       test_parse_int64.o
  CCLD     test_parse_int64
  CC       test_deep_copy.o
test_deep_copy.c: In function 'main':
test_deep_copy.c:201:14: error: unused variable 'special_output' [-Werror=unused-variable]
  const char *special_output = json_object_to_json_string(dest_with_serializer);
              ^~~~~~~~~~~~~~
test_deep_copy.c:198:8: error: unused variable 'dst_userdata' [-Werror=unused-variable]
  char *dst_userdata = json_object_get_userdata(dest_with_serializer);
        ^~~~~~~~~~~~
make[2]: *** [test_deep_copy.o] Error 1
make[2]: Leaving directory `/home/nwani/m3/conda-bld/json-c_1520910223423/work/tests'
make[1]: *** [check-am] Error 2
make[1]: Leaving directory `/home/nwani/m3/conda-bld/json-c_1520910223423/work/tests'
```

I think all all tests should have:
```
#ifdef NDEBUG
#/**/undef/**/ NDEBUG
#endif
```

But I couldn't find a header file common to only tests.

andy5995:
resolves #406

### DIFF
--- a/json.h
+++ b/json.h
@@ -35,4 +35,8 @@ extern "C" {
 }
 #endif
 
+#ifdef TEST
+#include "test_common.h"
+#endif
+
 #endif
--- a/json.h
+++ b/json.h
@@ -31,12 +31,12 @@ extern "C" {
 #include "json_object_iterator.h"
 #include "json_c_version.h"
 
-#ifdef __cplusplus
-}
-#endif
-
 #ifdef TEST
 #include "test_common.h"
+#endif
+
+#ifdef __cplusplus
+}
 #endif
 
 #endif
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1,5 +1,5 @@
 
-AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/tests
+AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/tests -DTEST
 
 LDADD= $(LIBJSON_LA)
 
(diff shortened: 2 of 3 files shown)


hawicz:
This doesn't look valid to do.  Not everything that includes json.h will be a test, not all tests include json.h, and littering the json.h with test-specific content protected by an overly generic define "TEST" seems like it's asking for trouble.
If you're ending up with NDEBUG defined for some reason, rather than trying to override that with an undef it'd be better to identify the root cause and fix *that*.
If there's a need for a test_common.h header anyway, well that could be fine, but then each of the tests should be modified to include it explicitly as the very first thing they do.
So, sorry, but I'm not going to merge this.

andy5995:
> This doesn't look valid to do. Not everything that includes json.h will be a test, not all tests include json.h, and littering the json.h with test-specific content protected by an overly generic define "TEST" seems like it's asking for trouble.

Oh, I get what you're saying. Thanks for the explanation.

> If you're ending up with NDEBUG defined for some reason, rather than trying to override that with an undef it'd be better to identify the root cause and fix _that_.

No problem for me, I was trying to address that ticket. :)

> So, sorry, but I'm not going to merge this.

No worries. I had my doubts about it, too.

Issue #498, opened 2019-07-24
double_serializer_test fails (with valgrind)


andy5995:
* JSON_VER: git
* OS: Debian 10

```
ERROR: valgrind found errors during execution:
FAIL test_double_serializer.test (exit status: 1)
```

Actually appears to be related to bash...? All other checks pass.

```
~/src/json-c/_build/tests$ valgrind --leak-check=full ./test_double_serializer 
==10195== Memcheck, a memory error detector
==10195== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==10195== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==10195== Command: ./test_double_serializer
==10195== 
==10198== 
==10198== HEAP SUMMARY:
==10198==     in use at exit: 51,962 bytes in 828 blocks
==10198==   total heap usage: 1,051 allocs, 223 frees, 78,776 bytes allocated
==10198== 
==10198== 11 bytes in 1 blocks are definitely lost in loss record 83 of 328
==10198==    at 0x483577F: malloc (vg_replace_malloc.c:299)
==10198==    by 0x195E8D: xmalloc (in /bin/bash)
==10198==    by 0x18F51A: set_default_locale (in /bin/bash)
==10198==    by 0x135EE6: main (in /bin/bash)
==10198== 
==10198== LEAK SUMMARY:
==10198==    definitely lost: 11 bytes in 1 blocks
==10198==    indirectly lost: 0 bytes in 0 blocks
==10198==      possibly lost: 0 bytes in 0 blocks
==10198==    still reachable: 51,951 bytes in 827 blocks
==10198==         suppressed: 0 bytes in 0 blocks
==10198== Reachable blocks (those to which a pointer was found) are not shown.
==10198== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==10198== 
==10198== For counts of detected and suppressed errors, rerun with: -v
==10198== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
==10197== 
==10197== HEAP SUMMARY:
==10197==     in use at exit: 51,600 bytes in 813 blocks
==10197==   total heap usage: 1,028 allocs, 215 frees, 74,418 bytes allocated
==10197== 
==10197== 11 bytes in 1 blocks are definitely lost in loss record 82 of 314
==10197==    at 0x483577F: malloc (vg_replace_malloc.c:299)
[...]

Issue #309, opened 2017-03-20
GCC 7 reports various -Wimplicit-fallthrough= errors


marxin:
GCC properly reports various places in code, where a case label falls through to another label:

json_object.c:678:6: warning: this statement may fall through [-Wimplicit-fallthrough=]
json_tokener.c:339:12: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:418:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:419:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:420:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:421:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:422:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:423:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:424:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:425:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:426:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:427:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
linkhash.c:428:15: warning: this statement may fall through [-Wimplicit-fallthrough=]

Can you please take a look?

hawicz:
I expect those locations need /* FALLTHRU */ comments added.  Pull requests are welcome.

PR #499, opened 2019-07-24
.travis.yml:test on more recent clang and gcc versions

andy5995:
(#498)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: cpp
-cache: ccache
 matrix:
 # gcc
   include:
@@ -78,8 +77,8 @@ before_install:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then
       eval "${MATRIX_EVAL}";
       if [ -n "$MATRIX_EVAL" ] && [ "$TRAVIS_COMPILER" != "clang" ]; then
-        sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 1E9377A2BA9EF27F;
-        sudo add-apt-repository -y ppa:ubuntu-toolchain-r/ppa;
+        sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 6B05F25D762E3157;
+        sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test;
         sudo apt-get update -qq;
         sudo apt-get install -y $CC;
       fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,12 +24,6 @@ matrix:
     - os: linux
       dist: bionic
       compiler: gcc
-      addons:
-        apt:
-          packages:
-            - valgrind
-            - cppcheck
-            - doxygen
       env: CHECK="true"
     - os: linux
       dist: bionic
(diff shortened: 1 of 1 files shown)


andy5995:
The travis test fails for this due to #498  though it looks like the condition I added is never executed because the shell is exiting after the error returned by `make check`

.travis.yml
```diff
@@ -28,5 +90,13 @@ script:
   - make
 
 after_success:
-  - make check
-  - if type cppcheck &> /dev/null ; then cppcheck --error-exitcode=1 --quiet *.h *.c tests/ ; fi
+  - if [ -n "$CHECK" ]; then
+      make distcheck;
+      result=$?;
+      if [ $result != 0 ]; then
```
andy5995: No idea why this isn't working. I've used it successfully on another repo....

.travis.yml
```diff
@@ -28,5 +90,13 @@ script:
   - make
 
 after_success:
-  - make check
-  - if type cppcheck &> /dev/null ; then cppcheck --error-exitcode=1 --quiet *.h *.c tests/ ; fi
+  - if [ -n "$CHECK" ]; then
+      make distcheck;
```
andy5995: On Travis, and on my own system, I see 3/21 checks that fail when make distcheck is run.

andy5995:
Test fails on gcc 7 and 8, but I don't see the problem with the statement, and I can't reproduce on my system using gcc8, or in docker containers running arch (gcc 9), Ubuntu (gcc7).

Maybe it wants

`((json_parse_int64(get_string_component(jso), &cint)) == 0)` ?

```
json_object.c: In function ‘json_object_get_int64’:
json_object.c:712:6: error: this statement may fall through [-Werror=implicit-fallthrough=]
   if (json_parse_int64(get_string_component(jso), &cint) == 0)
      ^
json_object.c:715:2: note: here
  default:
  ^~~~~~~
Makefile:602: recipe for target 'json_object.lo' failed
make[2]: *** [json_object.lo] Error 1
make[2]: Leaving directory '/home/travis/build/json-c/json-c'
Makefile:666: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/home/travis/build/json-c/json-c'
Makefile:464: recipe for target 'all' failed
make: *** [all] Error 2
```

andy5995:
> Maybe it wants
> 
> `((json_parse_int64(get_string_component(jso), &cint)) == 0)` ?
> 

Oh, I see now that I was completely wrong about what a "fallthrough" is.

.travis.yml
andy5995: > No idea why this isn't working. I've used it successfully on another repo....

Oh, my mistake; it was only displaying the logs because I had the condition wrong. Doesn't work. removed.

.travis.yml
andy5995: > On Travis, and on my own system, I see 3/21 checks that fail when make distcheck is run.

I fixed 2 of the tests in https://github.com/json-c/json-c/pull/500 , the 3rd one is still broken (https://github.com/json-c/json-c/issues/498)

andy5995:
I've rebased this so it includes https://github.com/json-c/json-c/commit/68abf12afab0ff40112e20ef59c8fd865f30c972

hawicz:
> 
> Oh, I see now that I was completely wrong about what a "fallthrough" is.

Per https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/ the default setting for this is -Wimplicit-fallthrough=3, and we went through fixing these errors at least once already (see issue #309), so I wonder what's different here?
Line 714 of json_object.c already has the FALLTHRU comment, which should suppress the warning.

.travis.yml
andy5995: 3rd one fixed by @hawicz in https://github.com/json-c/json-c/commit/68abf12afab0ff40112e20ef59c8fd865f30c972

andy5995:
> > Oh, I see now that I was completely wrong about what a "fallthrough" is.
> 
> Per https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/ the default setting for this is -Wimplicit-fallthrough=3, and we went through fixing these errors at least once already (see issue #309), so I wonder what's different here?

It was the ccache line at the top, now removed, build is passing.

https://stackoverflow.com/questions/45129741/gcc-7-wimplicit-fallthrough-warnings-and-portable-way-to-clear-them

andy5995:
Fails the distcheck, but after you merge #500, I'll rebase and it should pass.

andy5995:
rebased

andy5995:
I see Ubuntu bionic is available now on Travis. That would mean (I think) no need to add the toolchain ppa. I'm gonna do a little testing on another repo first though.

Unless you'd rather test on xenial, which is a little older. Any preference?

andy5995:
Yeah, bionic will work fine.

https://docs.travis-ci.com/user/reference/bionic/

Which would you like?

hawicz:
If only one, then let's go with Bionic.

andy5995:
> If only one, then let's go with Bionic.

Actually we can do more than one. I just pushed a change with the update. I kinda guessed at what you'd want.

andy5995:
I'll leave this as-is now for you to review. There's a bit of mix of bionic and xenial.

hawicz:
Looks good to me.  Thank you!

andy5995:
You're welcome @hawicz :)

PR #510, opened 2019-11-29
Bump so-name and improve CMake

besser82:
As stated in the individual commit messages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,13 @@ add_library(${PROJECT_NAME}
     ${JSON_C_HEADERS}
 )
 
+# Set the SO-VERSION and the build version for the built version.
+set_target_properties(${PROJECT_NAME}
+    PROPERTIES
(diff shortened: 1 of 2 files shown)


ploxiln:
I think that a feature addition typically merits an so minor bump. You said there were also backwards-incompatible ABI changes - it is probably worth mentioning at least one one significant example.

It is probably also worth adding a comment to the commit and to the cmake config about what CMP0075 is.

hawicz:
Also, as of https://github.com/json-c/json-c/blob/11a638048da0344a3bd5f38722ddecdf5ec22041/CMakeLists.txt I've included the VERSION and SOVERSION properties.

hawicz:
If I'm understanding CMP0075 correctly, it would make the target_link_libraries lines in tests/CMakeLists.txt irrelevant, no?

hawicz:
Not merging, since two of the changes have been made independently, and it's not clear that the library version needs to be bumped.

PR #511, opened 2019-12-11
Reduce lines for better optimization

hoaianhkhang:


hawicz:
"Optimizing" tests is rather pointless.  Also, your changes to use my_array_length are wrong, since you never recalculate that value.  Finally, this will fail to build on certain old compilers (mainly Visual Studio) due to variable declarations that aren't at the beginning of the functions.
Thanks for the suggestion, but I won't be merging this.

hoaianhkhang:
I see, may I ask is there anything else that I can do to help optimize the program, since I am new I do not really have experience yet.

hawicz:
I would ignore anything to do with the tests, they're not worth "optimizing".  If you're new to json-c, I recommend picking some task you might want to do that involves JSON, use the library to perform that task, whatever it might be, and if you find some deficiency in the library only *then* consider whether there's an improvement or bug fix that you can contribute.

PR #517, opened 2019-12-19
Lja mods

MiesSuomesta:
Some fixes that would be nice-to-have .. nothing is broken but bit better ?


hawicz:
Much of the changes here have been rendered irrelevant by the clang-format reformatting and the switch to cmake.
The stringdup() change is just broken and must not be included, plus it needlessly loses out on any platform-specific optimizations that strdup might have.
Changes like s/malloc/calloc/, and setting local variables to NULL just add pointless overhead and should not be done either.
However, the addition of some defines in printbuf.c (PRINTBUF_DEFAULT_SIZE, etc...) and the use of sizeof instead of hardcoded values could be worthwhile (though please avoid bad cases like "long unsigned int" when it should be "size_t").  If you want to re-spin that as a new PR I'll take a look at merging it.

PR #524, opened 2020-01-03
Increase coverage

dota17:


dota17:
For increasing the line coverage, I tested these special error scenarioes before, like the wrong value of ```json_type```, nonexistent file "./test_cast.test". emmmm, it seems a bit inappropriate and far-fetched now. I have removed these.

Issue #108, opened 2013-09-26
Decoding of string with null-byte.


remicollet:
String may contains null byte, ex "foo\u0000bar"

While json-c 0.11 allow to decode such value, I don't find an easy solution to decode such key in an object.

Ex : {"foo\u0000bar":"bar\u0000baz"}

Related to https://github.com/remicollet/pecl-json-c/issues/7

rgerhards:
Sorry for re-viving this old thread. I came accross the NUL byte problem when working on some performance enhancements. It looks like json-c as a producer supports them, so there is indeed some inconsistency. But what I wonder mostly is how would one expect to be able to work with them in a generic type of API? All APIs seem basically to work with C strings. So let's assume a NUL is read in. How would you expect to pass this buffer to a caller? Of course, you can pass buffer pointer and size, but is that what a typical C program expects? I would really appreciate some feedback on this issue.

hawicz:
Well, no, a C program written to work with "normal", nul terminated strings will not expect a buffer and size.  As I mentioned before, there would need to be "c_string" (i.e. the buffer+len data structure) variants of all of the json-c API functions.  If you wanted to take advantage of any features that allow for embedded nul characters your C code would need to change to use those new (not yet written) functions.

rgerhards:
Well, I have totally **no** desire to use NUL bytes unencoded in my program. But if we do not want to support that in json-c, we could officially state so. In this case, some simplifications could be done inside the code base (there is even a test for NUL byte encoding in the testbench...).

hawicz:
I think you misunderstand me.  I am not saying that json-c should not support that, I'm saying that it will be a significant change to fully do so.  I expect that adding c_string variants of all the API functions will actually mean changing *all* internal handling of strings to use c_string, and turning the current asciiz API calls to be wrappers around the c_string ones.  However, I haven't actually spent any time to evaluate the actual scope of the change, so who knows, maybe it'll be easy. :)
Clearly, something should be done, and if not the full conversion to c_string, then it would be a good idea to at least cause embedded \u0000's to result in an error, perhaps with "quick and dirty hack" options (as @skaes said) to either pass those through as-is, or re-enable the current rather broken behavior.

dota17:
About this issue, I want to know do you have one mature and preferred amendment. Can you give me some advice？

MikuChan03:
[...]

PR #528, opened 2020-01-10
Solve issue #108. Skip \u0000 while parsing.

dota17:
As discussed in [#108](https://github.com/json-c/json-c/issues/108).  \u0000 was skipped during token analysis.Did not escape  \u0000 to  \0.


hawicz:
Sorry, I can't merge this.  It's quite a hack that will break any existing usage that happens to handle json strings like:

    "x\\u0000x"

That'll turn into the 8 in-memory bytes: x, \, u, 0, 0, 0, 0, x
Which, with your changes, will then incorrectly be serialized into:

    "x\u0000x"

PR #531, opened 2020-01-10
validate utf-8 string  

dota17:
plan 2: validate utf-8 string  during parsing string.  #122 
Validate fuction has been added into macro `PEEK_CHAR(dest, tok)` .


dota17:
1. Add a new JSON_TOKENER_VALIDATE_UTF8 (default:0x10) flag and optimize the validate function.
2. performance testing: 



I test it with [nativejson-benchmark](https://github.com/miloyip/nativejson-benchmark) 10 times.  The conventional parse time without utf-8 validation is 105ms and the average parse time with utf-8 validation is 108.1ms. I think it's a acceptable result. @hawicz

PR #536, opened 2020-01-20
add json_object_new_null()

dota17:
issue #226

### DIFF
--- a/json_object.h
+++ b/json_object.h
@@ -965,6 +965,11 @@ JSON_EXPORT int json_object_set_string(json_object* obj, const char* new_value);
  */
 JSON_EXPORT int json_object_set_string_len(json_object* obj, const char* new_value, int len);
 
+/** Create a new empty json_object of type null and return NULL.
+ *
+ */
+JSON_EXPORT struct json_object* json_object_new_null();
+
 /** Check if two json_object's are equal
  *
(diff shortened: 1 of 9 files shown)


json_object.h
```diff
@@ -965,6 +965,11 @@ JSON_EXPORT int json_object_set_string(json_object* obj, const char* new_value);
  */
 JSON_EXPORT int json_object_set_string_len(json_object* obj, const char* new_value, int len);
 
+/** Create a new empty json_object of type null and return NULL.
```
hawicz: It doesn't create anything; this comment is misleading.  Let's have it just say something like 
"This method exists only to provide a complementary function along the lines of the other json_object_new_* functions.  It always returns NULL, and it is entirely acceptable to simply use NULL directly."

tests/test1.c
```diff
@@ -179,6 +179,9 @@ int main(int argc, char **argv)
 	printf("my_int=%d\n", json_object_get_int(my_int));
 	printf("my_int.to_string()=%s\n", json_object_to_json_string(my_int));
 
+	my_null = json_object_new_null();
```
hawicz: It seems a bit silly to write these tests pretending that we actually have a real object here, but as long as it works, ok.

PR #537, opened 2020-01-21
Replace '\0' only when parsing key, not change data in value.

dota17:
Before the hash value is calculated, the replacement "\ 0" operation is placed first, and then the parsing of the value will preserve the original logic. #528

### DIFF
--- a/json_tokener.c
+++ b/json_tokener.c
@@ -68,6 +68,8 @@ static const int json_true_str_len = sizeof(json_true_str) - 1;
 static const char json_false_str[] = "false";
 static const int json_false_str_len = sizeof(json_false_str) - 1;
 
+static char *replace_null_character(struct printbuf *puf);
+
 static const char* json_tokener_errors[] = {
   "success",
   "continue",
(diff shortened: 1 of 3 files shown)


hawicz:
I _really_ dislike these half-assed hacks to work around the problem of null bytes.  It's inefficient, and it just replaces one problem with a new one.  If you swap out a null byte with any kind of "magic value", you then introduce the problem of how to handle input with that magic value.  As you can see in your test output, the re-serialized value is still wrong, just in a different way.

Yes, the current behavior is broken, but at least it's broken in a way that is understandable to a C programmer, namely that an embedded null byte will cause truncation of a string.

If you want to do something to improve the situation, sketch out what would be needed to start passing a key length down to json_object_object_add, and from there down to the linkhash functions like lh_get_hash.  Exposing that in a public API would be quite intrusive, but perhaps we could at least get internal serialization code to handle it.

hawicz:
I won't be merging this PR.  See #715 for a potential alternate solution (though even that one needs another look before merging)

Issue #520, opened 2019-12-27
Adding type uint64


JayyCee:
I've enhanced the code with adding _uint64_ functions in my local branch.  This _uint64_ type is needed for my project that needs to handling a full data value of a 64-bit register from a hardware module.

I would like to share my enhancements.  What are the steps for me to do to make a pull-request?

--JC

hawicz:
This was asked before by @topilski , see https://github.com/json-c/json-c/issues/326
However, the implementation leaves a lot to be desired.  In particular, what he suggested adds a whole new "uint" type, but doesn't appear to do anything to deal with overflow of the internal signed storage field, and particular poor behavior when trying to retrieve negative values (i.e. just returns 0).

Also, any attempt to add additional bits is going to increase overhead and have various tradeoffs, and I'm skeptical that it's worth doing that for just a single bit.  On the other hand, if you want to bump things up to 128 bits, then things get more interesting.  Either way, I'd expect a characterization from you as to what you believe the benefits and drawbacks are/will be, and why it's justified to include the changes you want.

JayyCee:
I see where you come from regarding adding 64-bit thing.  

The single benefit:
Using current master's uint/int type, it gets a wrong value of -1 or the like,  when processing an unsigned integer value of 0xFFFFFFFFFFFFFFFF from an hardware register.

After adding  uint64_t capable of functions,  there is a working function of "json_object_new_uint64(0xFFFFFFFFFFFFFFFF), and then json_obj.to_string() function correctly produces a right number for it.
 
Trying the 128-bit thing is a very interesting challenge and it is beyond of my capability.  Hope you understand.

dota17:
Adding uint64 data type is complicated but not difficult. I have made it in #542 . UINT64 is useful and many json libraries have this feature. I am willing to promote the solution of this issue. 
@hawicz @JayyCee How do you think of #542? Need your suggestions.  0.0

hawicz:
Closing this, since #542 has been merged.  It adds support for uint64_t w/o adding a new json_type.  Overflow should be handled reasonably, by capping at UINT64_MAX, while negative values still end up being returned as 0.

Issue #326, opened 2017-05-25
Please extend api json_object_get_uint64


topilski:
Hi, can you add some usefull functions like `json_object_get_uint64 json_object_get_uint32`.

hawicz:
_uint32 would be fairly easy, just a matter of copying json_object_get_int() and replacing types and constants.  _uint64 is harder, since the internal value only parses and stores 63 bits of magnitude.  I'm not sure it's worth the trouble.

linghuazaii:
Yes, I think it is needed. I write a interface `json_object_get_uint64` for internal use. Waiting for this `new-feature`.

dota17:
Adding uint64 data type has been shown in #542 .You can review it and put forward some rectification opinions.

hawicz:
Closing, since support was added in #542.

Issue #143, opened 2014-07-18
RFE / enhancement for full 64-bit signed/unsigned support


dlethe:
I'm simply going to have to make some mods that facilitate handling 64 bit signed and unsigned numbers correctly so there is no loss of precision, especially if javascript is dealing with the numbers.  [For those not familiar with large numbers, javascript converts to float internally so numbers > 53 bits or so are rounded, so all math breaks unless you use special libraries]

Lots of ways to do this, but I hate to reinvent the wheel and get this incorporated into main repository.  

Seeing things like splitting into 2 x 32-bit ints [highorder,low order]
and twitters technique of sending large integers as:
 {"id": 10765432100123456789, "id_str": "10765432100123456789"}   instead of "id": 10765432100123456789 

To indicate the use. If numbers match then no worries, if not, then user knows it is a large integer so they have to work with the '_str' value and use some additional code to convert.

As incorporating this into the library itself touches a lot of code, want to run this by the community and  see if something like json_object_parse_split_uint64, split_int64, and associated tokening logic and such already exists, or if this is something best left out of it.

Thoughts?

hawicz:
#542 added support for uint64_t to json-c, but didn't do anything to address compatibility with javascript.  Defining a "special" serialization format seems like something that should probably sit on top of json-c, rather than implemented inside of it, so I'm closing this issue.

PR #542, opened 2020-02-25
add uint64 data to json-c

dota17:
Many people want to add uint64 data into json-c in #520, #326, #143. I believe it is necessary and useful, so I make it in this commit. Review please.

Added content:
1. Add ```uint64_t c_uint64``` data type and other related parameters, such as ```json_type_uint```.
2. Add new functions:
+ ```json_object_uint_to_json_string()```
+ ```json_object_new_uint64```
+ ```json_object_get_uint64()```
+ ```json_object_set_uint64()``` 
+ ```json_object_uint_inc()``` 
+ ```json_parse_uint64()``` 
3. Modify functions: (add uint64_t data/type )
+  ```json_object_set_serializer()```,
+ ```json_object_get_boolean()```
+ ```json_object_get_int()```
+ ```json_object_get_int64()```
+ ```json_object_get_double()```
+ ```json_object_equal()```
+ ```json_c_shallow_copy_default()```
4. Modify ```json_tokener_parse_ex()```function:
> if the number is double, just parse it as double. 
if the number is not double, then distinguish if it is negative. 
----if it is negative, parse it as int64 data.
----if it is not negative , parse it as uint64 data. Last, judge ```numuint64 <= INT64_MAX```, 
--------if yes, transfer it as int64 data.
5. Add related test cases.
```test_cast.c```, ```test_compare.c ```, ```test_int_add.c```, ```test_parse.c```, ```test_parse_int64.c```, ```test_set_value.c```.

As you can see, I have modified many files and achived adding-uint64.  Welcome your comments and suggestions

### CODE REVIEW

**Comment by hawicz** [comment]
Your changes look decent, but defining a whole new `json_type_uint` seems like an unnecessarily large breaking change, just to get one additional bit of magnitude.
Yes, that's a fairly straightforward way of adding this support, but I'm thinking that it might be a bit better to keep just a single `json_type_int` type exposed, and distinguish the actual width of the number only internally within the library.  e.g. `int64_t c_int64;` in the json_object structure would instead turn into something like:

    union data {
        ...
        struct {
            union {
                int64_t c_int64;
                uint64_t c_uint64;
            } u_int;
            int is_unsigned;
        } c_int;
        ...
    } o;

Which would allow things like json_object_get_type() to continue to return json_type_int, while also providing a easier way forward for future updates to expand the available magnitude beyond 64-bits (perhaps even arbitrary precision integers).

**Comment by dota17** [comment]
ok, I know your meanings. It's really a good solution, I will try to change it.

**Comment by hawicz** [comment]
Thanks.  I expect this will have the side-effect of making all places where we currently refer to c_int64 a bit more involved, but I'm hoping that ends up not being too bad.  If you're going to try this then I guess we'll see. :)

**Comment by dota17** [comment]
All this is a digression, could you open the coveralls right for json-c, ha-ha. It is a interesting tool.

File: json_object.c [hunk 1]
```diff
@@ -618,7 +623,10 @@ static int json_object_int_to_json_string(struct json_object* jso,
 {
 	/* room for 19 digits, the sign char, and a null term */
 	char sbuf[21];
-	snprintf(sbuf, sizeof(sbuf), "%" PRId64, jso->o.c_int64);
+	if (jso->o.c_int.cint_type == json_object_int_type_int64)
+	  snprintf(sbuf, sizeof(sbuf), "%" PRId64, jso->o.c_int.cint.c_int64);
+	else if (jso->o.c_int.cint_type == json_object_int_type_uint64)
```
hawicz: Let make this just a plain "else", to ensure we write *something* to sbuf.  Please adjust the other cint_type if conditions and switch statements similarly, to ensure we have a more definite action taken rather than just falling through.

json_object.c
```diff
@@ -675,9 +690,9 @@ int32_t json_object_get_int(const struct json_object *jso)
 }
 
 int json_object_set_int(struct json_object *jso,int new_value){
-	if (!jso || jso->o_type!=json_type_int)
+	if (!jso || jso->o_type!=json_type_int || jso->o.c_int.cint_type != json_object_int_type_int64)
```
hawicz: er.. only allowing int64 types through here is going to make for some non-obvious behavior.
Calling this should just switch cint_type to int64.

json_object.c
```diff
@@ -720,22 +754,76 @@ int64_t json_object_get_int64(const struct json_object *jso)
...
+			return cuint;
+		/* FALLTHRU */
+	default:
+		return 0;
+	}
+}
+
 int json_object_set_int64(struct json_object *jso,int64_t new_value){
-	if (!jso || jso->o_type!=json_type_int)
+	if (!jso || jso->o_type!=json_type_int || jso->o.c_int.cint_type != json_object_int_type_int64)
```
hawicz: As with _set_int(), this should set cint_type.
Also, let's avoid the dup code and have _set_int() call _set_int64(), since they do the exact same thing.

json_object.c
```diff
@@ -720,22 +754,76 @@ int64_t json_object_get_int64(const struct json_object *jso)
...
+		return 0;
+	jso->o.c_int.cint.c_int64=new_value;
+	return 1;
+}
+
+int json_object_set_uint64(struct json_object *jso,uint64_t new_value){
+	if (!jso || jso->o_type!=json_type_int || jso->o.c_int.cint_type != json_object_int_type_uint64)
```
hawicz: As with _set_int64(), this should just set cint_type instead of checking it.

json_object.c
```diff
@@ -720,22 +754,76 @@ int64_t json_object_get_int64(const struct json_object *jso)
...
 		return 0;
-	jso->o.c_int64=new_value;
+	jso->o.c_int.cint.c_uint64=new_value;
 	return 1;
 }
 
 int json_object_int_inc(struct json_object *jso, int64_t val) {
-	if (!jso || jso->o_type != json_type_int)
+	if (!jso || jso->o_type != json_type_int || jso->o.c_int.cint_type != json_object_int_type_int64)
```
hawicz: This shouldn't silently refuse to increment other cint_type.  Let's have it handle any cint_type, and if we happen to go past INT64_MAX, switch it to _type_uint64.

json_object.c
```diff
@@ -961,7 +1049,12 @@ double json_object_get_double(const struct json_object *jso)
   case json_type_double:
     return jso->o.c_double;
   case json_type_int:
-    return jso->o.c_int64;
+    switch(jso->o.c_int.cint_type) {
+		case json_object_int_type_int64:
```
hawicz: something went wonky with the whitespace here.

json_object.c
```diff
@@ -1340,7 +1433,14 @@ int json_object_equal(struct json_object* jso1, struct json_object* jso2)
 			return (jso1->o.c_double == jso2->o.c_double);
 
 		case json_type_int:
-			return (jso1->o.c_int64 == jso2->o.c_int64);
+			if ((jso1->o.c_int.cint_type == json_object_int_type_int64)
```
hawicz: int64 and uint64 are comparable.  If they have the same actual value, let's consider them equal regardless of any differences in cint_type.

json_object.h
```diff
@@ -181,6 +181,12 @@ typedef enum json_type {
   json_type_string
 } json_type;
 
+/* json object int type, support extension*/
+typedef enum json_object_int_type {
```
hawicz: I was wondering if you were going to create a enum like this. :)
Please move this to json_object_private.h

json_object.c
dota17: ok, done it

json_object.c
dota17: ok, done

json_object.c
dota17: ok, done

json_object.c
dota17: ok, done

json_object.c
dota17: ok, done. I have implemented the mutual conversion from int64 to uint64 in ```json_object_int_inc``` . I forgot this point before. It is necessary and shows superiorities of the new int data structure wihch you suggest.

json_object.c
dota17: ok, modified.

json_object.c
dota17: And, although you did not mark the ```json_object_uint_inc```, I have modified it as the same with ```json_object_int_inc```. 
> int64 +uint 64 ...

json_object.h
dota17: ok, done.

json_object.c
dota17: agree, done.

json_object.c
```diff
@@ -591,7 +591,14 @@ json_bool json_object_get_boolean(const struct json_object *jso)
 	case json_type_boolean:
 		return jso->o.c_boolean;
 	case json_type_int:
-		return (jso->o.c_int64 != 0);
+		switch(jso->o.c_int.cint_type) {
+		case json_object_int_type_int64:
+			return (jso->o.c_int.cint.c_int64 != 0);
+		case json_object_int_type_uint64:
+			return (jso->o.c_int.cint.c_uint64 != 0);
+		default:
```
hawicz: if we're going to have explicit default/else clauses for cint_type checks so catch unexpected cint_type values then they should lead to assert() calls, not an arbitrary value return. 
However, it'd also be ok (and avoid dead code) to have it be simply e.g.:

    switch (jso->o.c_int.cint_type) {
    case json_object_int_type_int64:
        return (jso->o.c_int.cint.c_int64 != 0);
    default: // i.e. json_object_int_type_uint64
        return (jso->o.c_int.cint.c_uint64 != 0);

dota17: I get it

json_object.c
```diff
@@ -720,24 +755,116 @@ int64_t json_object_get_int64(const struct json_object *jso)
...
+		return 0;
+	}
+}
+
+int json_object_uint_inc(struct json_object *jso, uint64_t val) {
+	if (!jso || jso->o_type != json_type_int)
+		return 0;
+	switch(jso->o.c_int.cint_type) {
+	case json_object_int_type_int64:
+		if ((uint64_t)jso->o.c_int.cint.c_uint64 + val> UINT64_MAX) {
```
hawicz: This can't ever be true.  There are all kinds of overflow pitfalls here.  Also, I'm not sure it's even worth having a separate json_object_uint_inc() at all, but even if it is I feel like it should be possible to use some common code between this and json_object_int_inc().

json_object.c
hawicz: Though you adjusted this location, there are several other places father below that could use similar treatment.

json_object.c
dota17: ha, It's my fault. Should be c_int64. There is some same errors here.

json_object.c
```diff
@@ -720,24 +755,116 @@ int64_t json_object_get_int64(const struct json_object *jso)
...
+		} else if (val < 0 && jso->o.c_int.cint.c_int64 < INT64_MIN - val) {
+			jso->o.c_int.cint.c_int64 = INT64_MIN;
+		} else {
+			jso->o.c_int.cint.c_int64 += val;
+		}
+		return 1;
+	case json_object_int_type_uint64:
+		if (val > 0 && jso->o.c_int.cint.c_uint64 > UINT64_MAX - (uint64_t)val) {
+			jso->o.c_int.cint.c_uint64 = UINT64_MAX;
+		} else if (val < 0 && jso->o.c_int.cint.c_uint64 < (uint64_t)(-val)) {
```
hawicz: I don't think "-val" will work when val == INT64_MIN.  -INT64_MIN > INT64_MAX, so it no longer fits in the variable.

json_object.c
dota17: Or delete json_object_uint_inc（）?

json_object.c
dota17: ok

json_object.c
```diff
@@ -720,24 +755,116 @@ int64_t json_object_get_int64(const struct json_object *jso)
...
+			jso->o.c_int.cint.c_int64 = (int64_t)jso->o.c_int.cint.c_uint64 + val;
+			jso->o.c_int.cint_type = json_object_int_type_int64;
+		} else if (val < 0 && jso->o.c_int.cint.c_uint64 >= (uint64_t)(-val)) {
+			jso->o.c_int.cint.c_uint64 = jso->o.c_int.cint.c_uint64 - (uint64_t)(-val);
```
hawicz: Slightly simpler would be just jso->o.c_int.cint.c_uint64 += val;

dota17: ok

json_object.c
dota17: ok

json_object.c
```diff
@@ -720,24 +755,116 @@ int64_t json_object_get_int64(const struct json_object *jso)
...
-	if (val > 0 && jso->o.c_int64 > INT64_MAX - val) {
-		jso->o.c_int64 = INT64_MAX;
-	} else if (val < 0 && jso->o.c_int64 < INT64_MIN - val) {
-		jso->o.c_int64 = INT64_MIN;
-	} else {
-		jso->o.c_int64 += val;
+	switch(jso->o.c_int.cint_type) {
+	case json_object_int_type_int64:
+		if (val > 0 && jso->o.c_int.cint.c_int64 > INT64_MAX - val) {
+			jso->o.c_int.cint.c_uint64 = (uint64_t)jso->o.c_int.cint.c_int64 + (uint64_t)val;
```
hawicz: Oh, neat, this can't overflow the unit64.  At most it'll get to INT64_MAX * 2 == UINT64_MAX - 1.  👍

dota17: yes, haha

json_object.c
hawicz: It's also likely to be error prone in that passing a negative value for val won't do what you might expect a first glance.
Let's drop this function, at least for now, and only consider adding it if someone has an actual need for it.

dota17: OK，I will delete it. If need it , I will add ```json_object_uint_inc()``` later.

hawicz:
I think this could still use a couple more tweaks (i.e. improved assert calls, a bit of simplification in object_equal), but it's good enough to merge.  I'll make those other changes later today.

dota17:
ok, thx.

hawicz:
Done (commit  b3296e7)

dota17:
good，It's improved a lot. And #520 , #326 , #143 maybe should be closed.

andrewhodel:
Can't you just input an int64 then cast it as unsigned when you read it?

hawicz:
Input where, in your json input?  That means you'd have things like:

    { "foo": -9223372036854775808,
       "bar":  -1 }

when you really mean:

    { "foo": 9223372036854775808,
       "bar": 18446744073709551615 }

Which would be quite confusing, not to mention not at all interoperable with any other json implementation that might support numbers that large.

andrewhodel:
Sure, but you might need to support version 0.12 not 0.15



Andrew

PR #546, opened 2020-02-27
Add uninstall target in cmake

dota17:
Add uninstall target in cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,6 +335,12 @@ else (DOXYGEN_FOUND)
 	message("Doxygen need to be installed to generate the doxygen documentation")
 endif(DOXYGEN_FOUND)
 
+# uninstall
+add_custom_target(uninstall
+  COMMAND cat ${PROJECT_BINARY_DIR}/install_manifest.txt | xargs rm -rf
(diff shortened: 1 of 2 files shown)


dota17:
hi, hawicz
I have added several cmake features before. What else do you think needs to be added in cmake?

hawicz:
er... that's odd, I could have sworn I had commented on this change.
Anyway: using the -r option to rm is unnecessary and dangerous.  The contents of install_manifest.txt look like it's just files, not directories, and even if it were directories you wouldn't want to remove them if they happened to contain non-json-c files in them.

dota17:
I have modified the make_uninstall in CMakeLists.txt and added the cmake/cmake_uninstall.cmake.in file according the [official manual](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake)

hawicz:
oof, that's a whole pile of complexity for a marginally useful feature.  :(  Sure, they might recommend that for "cross platform compatibility", but let's not drink _all_ the kool aid.
Please feed the contents of install_manifest to rm like you were doing before, just without the "-r" option to recurse.

dota17:
ok, finished

PR #547, opened 2020-02-29
modify json-c default build type, and fix up the assert() errors in t…

dota17:
1. Solve the problemof testcase failure during cmake release-building. 
2. Modify json-c default cmake build type.
3. Add the default compilation optimization options.
See details in #548

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ if(NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE debug)
 endif()
 
-set(CMAKE_C_FLAGS_RELEASE   "${CMAKE_C_FLAGS_RELAESE} -O2")
+set(CMAKE_C_FLAGS_RELEASE   "${CMAKE_C_FLAGS_RELEASE} -O2")
 
 # Include file check macros honor CMAKE_REQUIRED_LIBRARIES
 # i.e. the check_include_file() calls will include -lm when checking.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,8 @@ if(NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE debug)
 endif()
 
+set(CMAKE_C_FLAGS_RELEASE   "${CMAKE_C_FLAGS_RELAESE} -O2")
+
 # Include file check macros honor CMAKE_REQUIRED_LIBRARIES
 # i.e. the check_include_file() calls will include -lm when checking.
 if(POLICY CMP0075)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -238,6 +243,9 @@ elseif ("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
     set(CMAKE_C_FLAGS           "${CMAKE_C_FLAGS} /wd4701")
 endif()
 
+set(CMAKE_C_FLAGS_DEBUG     "${CMAKE_C_FLAGS_DEBUG} -O0")
```
hawicz: The problem with doing this is that it disables a lot of warnings that depend on the optimizer being enabled. :(

tests/test_deep_copy.c
```diff
@@ -2,7 +2,12 @@
 #include <stdlib.h>
 #include <stddef.h>
 #include <string.h>
+#ifdef NDEBUG
+#undef NDEBUG
 #include <assert.h>
+#else
```
hawicz: Why not put the assert.h after the undef NDEBUG, so you don't need duplication nor a #else ?

CMakeLists.txt
```diff
@@ -238,6 +243,9 @@ elseif ("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
...
+set(CMAKE_C_FLAGS_Release   "${CMAKE_C_FLAGS_Release} -O2")
```
hawicz: er... -O2 should have been be part of the compile flags already.  oops.  Let's set this, but on by default and for release builds.  If we need a non-optimized debug build, add something like "DebugNoOpt" to CMAKE_CONFIGURATION_TYPES that switches to -O0.
Also, per https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html, the variables end up all upper-case, i.e. CMAKE_C_FLAGS_RELEASE

dota17: ha, I tested the debug ```-O0```/default non```-O0``` and found the effect is basically the same. They are both have no compilation optimization. So we do not need to set CMAKE_C_FLAGS_DEBUG -O0.

tests/test_deep_copy.c
dota17: My fault, commit it too fast without thinking, 0.0

CMakeLists.txt
```diff
@@ -19,6 +19,13 @@ if(POLICY CMP0054)
     cmake_policy(SET CMP0054 NEW)
 endif()
 
+# set default build type if not specified by user
+if(NOT CMAKE_BUILD_TYPE)
+	set(CMAKE_BUILD_TYPE debug)
+endif()
+
+set(CMAKE_C_FLAGS_RELEASE   "${CMAKE_C_FLAGS_RELAESE} -O2")
```
ploxiln:
```suggestion
set(CMAKE_C_FLAGS_RELEASE   "${CMAKE_C_FLAGS_RELEASE} -O2")
```

PR #551, opened 2020-03-10
Allow hexadecimal number format convention parsing

nitzancarmi:
Allow parsing of the hexadecimal number standard
format (e.g. `0x<number>`).

Signed-off-by: Nitzan Carmi <nitzanc@mellanox.com>

### DIFF
--- a/json_tokener.c
+++ b/json_tokener.c
@@ -242,6 +242,30 @@ struct json_object* json_tokener_parse_verbose(const char *str,
 /* End optimization macro defs */
 
 
+int json_tokener_number_has_hex_prefix(const char *number)
+{
+	char *ptr_x;
+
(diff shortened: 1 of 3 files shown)


dota17:
1,  Standard json do not support octal and hexadecimal number. 
```A number is very much like a C or Java number, except that the octal and hexadecimal formats are not used.``` https://www.json.org/
2, Maybe you should add the test cases to verify your function implementation.

hawicz:
Sorry, this is totally out of the spec we're aiming for (RFC 7159), and not even in the newer one (RFC 8259), so I don't expect that we'll merge this.

nitzancarmi:
Alright then. Tnx!

PR #555, opened 2020-03-14
Format json-c with clang-format tool

dota17:
1. Add several kinds of Short Blocks/SingleLine option.
>AllowShortBlocksOnASingleLine:  true
AllowShortCaseLabelsOnASingleLine: true
AllowShortFunctionsOnASingleLine:  Empty
AllowShortIfStatementsOnASingleLine: true

2. The indent effects of ```IndentWidth``` and ```TabWidth```  will overlay, so do not add ```TabWidth```. 
> IndentWidth: 8
UseTab: ForIndentation

3. ```BreakBeforeBraces: Custom```. This is because we should not wrap the extern blocks. like:
```c
// in arraylist.h
#ifdef __cplusplus
extern "C" {
#endif
```
4. Partial comments in functions need to be adjusted manually. like:
```c
{
    /*      (formatted)
     *      (adjusted)
     *      (adjusted)
     */     (adjusted)
}
```
5. ```random_seed.c``` is not suitable for clang-format. So exclude it from the range of this format temporarily.

In a word, the codes of json-c have been well formatted. If we need to add or modify optimization options, we can format the codes on this basis.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+BasedOnStyle:  LLVM
+DerivePointerAlignment: false
+PointerAlignment: Left
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,53 @@
 BasedOnStyle:  LLVM
-DerivePointerAlignment: false
-PointerAlignment: Left
+# If true, clang-format will attempt to re-flow comments
+ReflowComments: false
+# The column limit.
+ColumnLimit: 100
+# Indent width for line continuations.
+ContinuationIndentWidth: 4
+# The number of columns to use for indentation.
+IndentWidth: 8
+# The number of columns used for tab stops.
+TabWidth: 8
+UseTab: ForIndentation
+
+# Options for aligning backslashes in escaped newlines.
+AlignEscapedNewlines: Left
+# Short Block Style
+AllowShortBlocksOnASingleLine:  true
+# If true, short case labels will be contracted to a single line.
+AllowShortCaseLabelsOnASingleLine: true
+# Dependent on the value, int f() { return 0; } can be put on a single line.
+AllowShortFunctionsOnASingleLine:  Empty
+# The brace breaking style to use.
+BreakBeforeBraces: Custom
+# Control of individual brace wrapping cases.
+BraceWrapping:
+  # Wrap class definition.
+  AfterClass: true
+  # Wrap control statements
+  AfterControlStatement: true
+  # Wrap enum definitions.
+  AfterEnum: true
+  # Wrap function definitions.
+  AfterFunction: true
+  # Wrap namespace definitions.
+  AfterNamespace: true
+  # Wrap struct definitions.
+  AfterStruct: true
+  # Wrap union definitions.
+  AfterUnion: true
+  # Wrap extern blocks.
+  AfterExternBlock: false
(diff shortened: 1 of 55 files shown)


hawicz:
Does changing the value used for IndentWidth have an effect?   I suspect it doesn't but let's make it consistent with ContinuationIndentWidth (i.e. set it to 4) unless that screws something up.

It's too bad we can't just tweak one of the BraceWrapping presets, but oh well.

Let's turn off AllowShortIfStatementsOnASingleLine, it seems to make a lot of the codeflow look inconsistent and IMO somewhat harder to follow.

There are a number of places where that shouldn't be reformatted, please add `/* clang-format off */` and `/* clang-format on */` comments around these sections:

- Arrays of constants (e.g. json_type_name) really shouldn't be merged into a single line.
- The mix and final macros in linkhash.c shouldn't be re-formatted, neither should hashlittle()
- errno_list[] in strerror_override.c
- probably others, I haven't had a chance to read through the full diff yet.

dota17:
ok, I will modify these optimazation options and block the partial content .  But if the ```IndentWidth``` is set as 4 and ```UseTab``` is set as ForIndentation, things will get worse and I test it many times. Do you mind I set the ```IndentWidth``` and ```ContinuationIndentWidth``` both are set as 8?

hawicz:
Oh, I see, it defaults to calculating tabs as 8 characters wide.  If you set IndentWidth and TabWidth both to 4, then things end up mostly the same, though it wraps lines a bit less often because it calculates the location of the characters after the tab differently.
So, let's keep IndentWidth:8, but please add an explicit TabWidth:8 line.

ContinuationIndentWidth uses spaces, not tabs, to indent the continuation lines (even when indenting 8 characters or more), so let's leave that as ContinuationIndentWidth:4.

dota17:
I think the overall clang-formatting effect of this time is pretty good. You may not have the enough time to read the entire file. I add the following key points:
1. Some statements in ```json_tokener.c``` are folded(indented). The numbers of nesting levels of these statements are too big. But I think the formatted results are acceptable. And if needed,  we can also set the ```ColumnLimit``` as 120 to alleviate the situation.
2. The ```hashlittle()``` in ```linkhash.c``` is adjusted/formatted manully and  the disabling formatting comments (```/* clang-format off/on */``` ) is added. Its original basic format is retained. And the other places where that shouldn't be reformatted get handled as the same way. Manual adjustment details can be seen in commit [2d3fc02](https://github.com/json-c/json-c/pull/555/commits/2d3fc028b718de7d440eb54a122891f4611e1bdf)
3. ```random_seed.c``` has not been formatted. Because code format affects function implementation. This file should be handled separately. I will deal with it later(maybe not this PR).
4. Is the new format of macros in debug.h suitable？I think it is well.

dota17:
hi @hawicz, Are you satisfied with the above formatting results. I think the overall effect is acceptable and I can adjust the format at any time again.

hawicz:
Yes, that all looks good.  Looks like this needs to be updated one more time to address the recent conflicts that appeared, but once that's done let's go ahead and merge it.

dota17:
ok. The coveralls  error should be related to setting of  ```COVERAGE DECREASE THRESHOLD FOR FAILURE```. I have set it as 5.0%.

PR #556, opened 2020-03-15
Fixes various Wreturn-type and Wimplicit-fallthrough errors on Mingw-w64

Jehan:
This is a recent regression since commit
6359b798479d379a3202e02c6a938d9b40c0d856 which added various assert(0)
calls (often replacing return-s).
With Ming-W64 compiler, json-c build was failing with various errors of
the sort:

> /home/jehan/dev/src/json-c/json_object.c: In function 'json_object_int_inc':
> /home/jehan/dev/src/json-c/json_object.c:841:1: error: control reaches end of non-void function [-Werror=return-type]
>   841 | }
>       | ^
> In file included from /home/jehan/dev/src/json-c/json_object.c:17:
> /home/jehan/dev/src/json-c/json_object.c: In function 'json_object_get_double':
> /home/jehan/.local/share/crossroad/roads/w64/json-c/include/assert.h:76:4: error: this statement may fall through [-Werror=implicit-fallthrough=]
>    76 |   (_assert(#_Expression,__FILE__,__LINE__),0))
>       |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jehan/dev/src/json-c/json_object.c:1070:7: note: in expansion of macro 'assert'
>  1070 |       assert(0);
>       |       ^~~~~~
> /home/jehan/dev/src/json-c/json_object.c:1072:3: note: here
>  1072 |   case json_type_boolean:
>       |   ^~~~

The problem is that Mingw-w64 does not consider assert() as a noreturn
(even assert(0)), because it has to be compatible by Microsoft
libraries. See the discussion here:
https://sourceforge.net/p/mingw-w64/bugs/306/

As a consequence, we still need to break or return values when relevant.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -597,7 +597,7 @@ json_bool json_object_get_boolean(const struct json_object *jso)
 		case json_object_int_type_uint64:
 			return (jso->o.c_int.cint.c_uint64 != 0);
 		default:
-			assert(!"invalid cint_type");
+			json_abort("invalid cint_type");
 		}
 	case json_type_double:
 		return (jso->o.c_double != 0);
@@ -734,7 +734,7 @@ int64_t json_object_get_int64(const struct json_object *jso)
 				return INT64_MAX;
 			return (int64_t)jso->o.c_int.cint.c_uint64;
 		default:
-			assert(!"invalid cint_type");
+			json_abort("invalid cint_type");
 		}
 	case json_type_double:
 		// INT64_MAX can't be exactly represented as a double
@@ -772,7 +772,7 @@ uint64_t json_object_get_uint64(const struct json_object *jso)
 		case json_object_int_type_uint64:
 			return jso->o.c_int.cint.c_uint64;
 		default:
-			assert(!"invalid cint_type");
+			json_abort("invalid cint_type");
 		}
 	case json_type_double:
 		// UINT64_MAX can't be exactly represented as a double
@@ -836,7 +836,7 @@ int json_object_int_inc(struct json_object *jso, int64_t val) {
 		}
 		return 1;
 	default:
-		assert(!"invalid cint_type");
+		json_abort("invalid cint_type");
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Jehan:
Hi!

We are building json-c on the GIMP CI which is broken (for Windows jobs) since commit 6359b798479d379a3202e02c6a938d9b40c0d856.

I was wondering though why the cross-compiler (Mingw-w64) was not seeing the `assert(!"invalid cint_type");` calls as noreturn anyway and doing the smart thing, so I searched a bit and found this discussion on the compiler mailing list: https://sourceforge.net/p/mingw-w64/bugs/306/

So apparently this is not really a bug and we cannot do a thing about it since this is necessary to stay compatible with Windows libraries. All we can do is be more thorough in our use of `assert()`, even if it seems redundant to add a `return` or a `break` after.

Thanks!

hawicz:
Rather than "return 0" after the assert, it'd be better to call abort(), or maybe even define an "abort_with_message" function, marked as no-return, that does a fprintf+abort.

Jehan:
> Rather than "return 0" after the assert, it'd be better to call abort(), or maybe even define an "abort_with_message" function, marked as no-return, that does a fprintf+abort.

It is indeed cleaner than misusing `assert()`.

I created `json_abort()` and added it in `json_utils.[ch]`. I wondered if you prefered 2 commits (one just for the new function and one for fixing the s/assert/json_abort/ usage), for a more organized log. Tell me if you prefer it, and I'll split the commit if so.

hawicz:
A single commit is fine, but you'll need to adjust that to work on windows, as Visual Studio doesn't understand the gcc __attribute__ syntax.  
https://ci.appveyor.com/project/hawicz/json-c/builds/31502300/job/03ig0yw9y64254ca

Jehan:
> A single commit is fine, but you'll need to adjust that to work on windows, as Visual Studio doesn't understand the gcc attribute syntax.

Ok I pushed a new version, with hopefully the right format (I don't have access to a Visual Studio).

hawicz:
Thanks!

PR #558, opened 2020-03-19
modify the doc target message

dota17:


dota17:
Simple message modification, I will merge it.

PR #563, opened 2020-04-02
Changed order of calloc args to match stdlib

robybeen:
Although it is currently working, it's worth to stick with the stdlib definition to avoid further problems


hawicz:
eh... "further problems" seem quite unlikely, but I guess if you're on a system that cares about alignment, and that won't give you a maximally aligned chunk of memory when the second argument to calloc is small, then perhaps this might help.  Doesn't hurt, so I'll merge it.

PR #566, opened 2020-04-07
add the json_types.h to Makefile.am

dota17:
When I update dota17/json-c, I found the CI didn't pass. But I do not know why the CI passed in  json-c/json-c master branch. Maybe the CI  do not finish? Coveralls may also be an influencing factor. I close the option ```USE STATUS API?``` in Coveralls. You can hava a try. @hawicz

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,6 +29,7 @@ libjson_cinclude_HEADERS = \
 	json_object_iterator.h \
 	json_pointer.h \
 	json_tokener.h \
+	json_types.h \
 	json_util.h \
 	json_visit.h \
 	linkhash.h \
--- a/vasprintf_compat.h
+++ b/vasprintf_compat.h
@@ -8,8 +8,6 @@
 
 #include "snprintf_compat.h"
 
-#include <malloc.h>
-
 #if !defined(HAVE_VASPRINTF)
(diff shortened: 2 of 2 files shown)


dota17:
It seems to be a problem with travis-CI.  The CI didn't pass and the travis-CI didn't report the error massage to json-c/json-c.  
https://travis-ci.org/github/json-c/json-c/builds/671657455?utm_medium=notification&utm_source=github_status

hawicz:
Why are you removing malloc.h?
Also, rather than changing Makefile.am we should fix the travis build to use cmake.

dota17:
agree，I just do this for passing CI. And I am trying changing cmake.

hawicz:
I just pushed an update to .travis.yml, let's see if it fixes it.

hawicz:
Nope, looks like I need to use a separate builddir for the distcheck target to work.

dota17:
Do we need to write and upload a file like the ```issues_closed_for_0.13.md```? Or just change the changelog.

dota17:
oh, I see the ```Notes for v0.14 release```

hawicz:
The travis build is now working w/ cmake, and once we confirm cmake has all the same features that configure did, all of the autoconf stuff is going away, so I'm closing this without merging.

PR #567, opened 2020-04-08
Install json_types.h with autotools build as well.

Jehan:
Since a recent commit (19bbf2c), this header must also be installed.
Otherwise any software building against json-c fails the build.


dota17:
json-c will release a new version. This version will eliminate all the autoconf machinery and use cmake to compiler.

hawicz:
Makefile.am and all of the autoconf machinery is gone now.  Please use cmake instead.

PR #568, opened 2020-04-08
Adding better support to MinGW

soyouzpanda:
Adding better support to MinGW by detecting if MinGW is used to avoid MinGW-make to crash:
There was an error caused by the two lines with MinGW:
`#define strtoll _strtoi64`
`#define strtoull _strtoui64`

image

### DIFF
--- a/json_util.c
+++ b/json_util.c
@@ -39,7 +39,7 @@
 #endif /* HAVE_UNISTD_H */
 
 #ifdef WIN32
-#if MSC_VER < 1800
+#if (MSC_VER < 1800) && !defined(__MINGW32__)
 /* strtoll/strtoull is available only since Visual Studio 2013 */
 #define strtoll _strtoi64
 #define strtoull _strtoui64


hawicz:
It looks like json_util.c is using the wrong approach to decide whether to use those defines.  It should be using "json_c_strtoll" and "json_c_strtoull" which are determined at cmake time.
Care to adjust your change to drop the defines from json_util.c entirely, and do a search+replace on the rest of the file?

soyouzpanda:
I just think by using MinGW, MSC_VER is undefined and he is treating it like MSC_VER is less than 1800. But MinGW already have strtoll and strtoull defined, so we don't need to define them. But he is defining them.
However, I tried what you said, but it doesn't work.
My original changes work perfectly with MinGW now.

hawicz:
Doesn't work _how_?  Are json_c_strtoll and json_c_strtoull not actually ending up defined?  Is the cmake handling defining them incorrectly?

hawicz:
Oh, wait a sec, the section in CMakeLists.txt is inside a "if (MSVC)" block, when it should just always be used.

hawicz:
hang tight, I'm fixing it...

hawicz:
Closing without pull, since the change here is no longer relevant.

PR #571, opened 2020-04-10
CMake: Bump SONAME to 5.

besser82:
The soname of libjson-c needs to be bumped as there were API changes introduced since the last release:

  * lh_abort() has been removed
  * lh_table_lookup() has been removed
  * TRUE and FALSE defines have been removed


hawicz:
Thanks.  We'll do this as part of the process for the upcoming 0.14 release.  (https://github.com/json-c/json-c/blob/95737df/RELEASE_CHECKLIST.txt#L48)

Issue #569, opened 2020-04-09
Handling of -Bsymbolic-function in CMakeLists.txt is deficient


hawicz:
The old autoconf machinery had an actual link test to make sure that the -Bsymbolic-function complier argument actually worked.  CMakeLists.txt is missing this, and thus building on systems that don't support that, such as osx, fails unless -DDISABLE_BSYMBOLIC is passed to cmake.

Once this is fixed, the special case in .travis.yml could be cleaned up as well.

dota17:
like this?
```
set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-Bsymbolic -Wl,-Bsymbolic-functions")
set (CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,-Bsymbolic -Wl,-Bsymbolic-functions")
```
https://github.com/apitrace/apitrace/blob/master/wrappers/CMakeLists.txt

hawicz:
No, we already set -Bsymbolic-functions, the problem is that we do it unconditionally without checking whether it is supported by the linker.

hawicz:
Fixed by @besser82 in PR #572

PR #572, opened 2020-04-10
Small fixes to CMakeLists

besser82:
* Fix appending of linker flags.
* Build Doxygen documentation out-of-tree.
* Check whether BSYMBOLIC is supported by the linker.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,7 +270,20 @@ if (NOT ("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC"))
 		add_compile_options("-D_REENTRANT")
 	endif()
 
-	if (DISABLE_BSYMBOLIC STREQUAL "OFF")
+	# OSX Mach-O doesn't support linking with '-Bsymbolic-functions'.
+	# Others may not support it, too.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -270,8 +270,9 @@ if (NOT ("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC"))
 		add_compile_options("-D_REENTRANT")
 	endif()
 
-	if (DISABLE_BSYMBOLIC STREQUAL "OFF")
-		set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKED_FLAGS} -Wl,-Bsymbolic-functions")
+	# OSX Mach-O doesn't support linking with '-Bsymbolic-functions'.
+	if (DISABLE_BSYMBOLIC STREQUAL "OFF" AND NOT APPLE)
```
hawicz, line 274: I'd really rather not hardcode a particular vendor like this.  If we're trying to use a linker flag it would be better to actually check whether it works (#569).

besser82: Fixed.

hawicz:
Awesome, thanks!

PR #575, opened 2020-04-11
Small cosmetics.

besser82:
* Fix line terminators to be UNIX.
* Clean trailing white-space.
* gitignore: Add build folder.
* Remove multiple trailing newlines at EOF.


dota17:
```
Fix line terminators to be UNIX.
```
yes, I faced the problem like, ```syntax error: unexpected end of file```, before.

PR #584, opened 2020-04-18
CMake: SOVERSION and the major library VERSION need to be in lockstep.

besser82:
The SONAME and the major library VERSION always need to be in lockstep.
The major library version MUST be bumped, when an interface is removed from the API.  If new interfaces are added to the API, the library's revision shall be bumped, but MUST be reset to 0 on an event that requires a bump of the major library version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,15 @@ if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING AND
 add_subdirectory(tests)
 endif()
 
+# Library version and SONAME.
+# SOVERSION and the major library VERSION need to be in lockstep.
+# Major library version: Bump this, when interfaces are removed
+# from the API.
+set(LIB_VERSION_MAJOR    4)
+# Library revision: Bump this, when new interfaces are added to
+# the API.  Reset to 0, in case of bumping the major version!
+set(LIB_VERSION_REVISION 0)
+
 # Set some packaging variables.
 set(CPACK_PACKAGE_NAME              "${PROJECT_NAME}")
 set(CPACK_PACKAGE_VERSION_MAJOR     "${PROJECT_VERSION_MAJOR}")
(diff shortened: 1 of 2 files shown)


besser82:
@hawicz As the release of 0.14.0 in rather near, the major library version should be bumped as well, since there are backwards incompatible changes of the public API.

Shall I add that to this PR so it will not be forgotten?

ploxiln:
thanks for preparing the report: https://besser82.fedorapeople.org/json-c-api/

It seems to me that the changes are to internals that are not intended to be directly used externally, and probably pretty much always are not so used. (But I suppose techniques could have been used to "hide" them better.)

Removed functions:
  * `lh_abort ( char const* msg, ... )`
  * `lh_table_lookup ( struct lh_table* t, void const* k )`

from `linkhash.h` which is "not recommended for direct use" and probably rarely used by consumers of json-c

Changed structs:
  * `struct json_object`

This struct is defined in `json_object_private.h` and there are no functions which would help you initialize one which you allocated on the stack or as a global, you pretty much only get these on the heap created by json-c functions.

Removed macros:
  * `#define | FALSE   ((json_bool)0)`
  * `#define | TRUE   ((json_bool)1)`

This technically makes the new version API incompatible, but the ABI is not affected.

Changed enums:
  * `enum json_tokener_error`

Just the the `json_tokener_error_size` was incremented, which is a trick used to automatically determine the max value of the enum, mostly for internal purposes, usually the consumer of json-c would just pass it to `json_tokener_error_desc()` ...

-----

Agreed, not enough internal details were hidden to give an ironclad guarantee of perfect ABI compatibility, and it's easy to just bump the SOVERSION. But I think that just bumping the "REVISION" would also be quite reasonable.

besser82:
@ploxiln Well, at least we have clashing symbols with `cryptsetup` (which uses json-c):

https://gitlab.com/cryptsetup/cryptsetup/-/commit/e6a356974330e3ae21579a5737976e9a2aad1b51

***

From my POV we should bump the soname if there is still a (vague) possibility of ABI incompatibility.

besser82:
Besides that the core library of the `dnf` package-manager in Fedora is a consumer of `json-c`, I would really prefer to go the safe way…

hawicz:
soname/version bumps happen at release time, and I _will_ be bumping the major version as part of the 0.14 release.

AFACT, "LIB_VERSION_MAJOR" isn't anything special, it's just another variable.  Perhaps you're thinking of "JSON_C_VERSION_MAJOR" to match the project name set by the project() call?  I'm referencing https://cmake.org/cmake/help/v3.17/manual/cmake-variables.7.html

In any case it wouldn't make any sense to use the project version as the soversion, so I don't understand what you're trying to make "in lockstep" here.  Can you explain?

hawicz:
I don't think there's anything to be done here.  If you think there's still a fix that needs to be done, feel free to reopen.

PR #585, opened 2020-04-20
CMake: Do not install config.h, as it is not a public header file.

besser82:
Installing this file will lead to clashes with applications, that link against json-c and include their own build-time config.h.
Anyways, that particular header file is not needed to build and link against json-c.

Likewise any public headers with generic names should be prefixed with `json_c` to avoid possible undesired clashes.

***

Please backport these commits to 0.14 as the current release may break building any apllication randomly for no obvious reason for this header file being present.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,7 +314,7 @@ set(JSON_C_PUBLIC_HEADERS
     ${PROJECT_SOURCE_DIR}/json_util.h
     ${PROJECT_SOURCE_DIR}/json_visit.h
     ${PROJECT_SOURCE_DIR}/linkhash.h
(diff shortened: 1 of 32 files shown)


hawicz:
Excluding config.h is fine, but we can't do this with arraylist.h, printbuf.h nor linkhash.h.  For better or for worse, they're part of the public API and have been for some time now.  This would be a good item to put on the todo list for a major 1.0 release though.

hawicz:
Added item to https://github.com/json-c/json-c/wiki/Proposed-changes-for-new-v1.0-version

hawicz:
config.h is now excluded from install on master and the json-c-0.14 branch.  Since that's the best we can do at the moment, I'm closing this.

PR #587, opened 2020-04-28
Store the hashValue to avoid repeating the hash calculation during the hash resize.

dota17:
Discuss:
Cache the hash value in lh_entry. This value is used directly during hash resize to calculate the index of the data in hash table. This method will reduce the times of hash calculations. The improving effect of the parse speed is more obvious in the case of multiple objects. But this method will add 4 bytes overhead for every lh_entry in 32-bit systems.

|json file|ave_time before/ms|ave_time after/ms|optimization ratio/%|
| ---- | ---- | ---- | ---- |
|[twitter.json](https://github.com/dota17/nativejson-benchmark/blob/master/data/twitter.json)|11.66|10.53|9.69|
|jeopardy_50m.json|964|920|4.5|
|[citm_catalog_1.6m.json](https://github.com/dota17/nativejson-benchmark/blob/master/data/citm_catalog.json)|28.82|27|6.3|

### DIFF
--- a/linkhash.c
+++ b/linkhash.c
@@ -540,7 +540,7 @@ int lh_table_resize(struct lh_table *t, int new_size)
 
 	for (ent = t->head; ent != NULL; ent = ent->next)
 	{
-		unsigned long h = lh_get_hash(new_t, ent->k);
+		unsigned long h = ent->hash;
 		unsigned int opts = 0;
 		if (ent->k_is_constant)
 			opts = JSON_C_OBJECT_KEY_IS_CONSTANT;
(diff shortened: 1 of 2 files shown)


hawicz:
This also increases the size of struct lh_entry, and is an ABI breaking change since you're re-ordering the fields, so we can't merge this as-is.  
Even if you were to put the new lh_entry->hash field at the end of lh_entry, it's still an externally visible change since the lh_table->table array is exposed.  External code _shouldn't_ be access that directly, but that can't be guaranteed.
Given that, I don't think we can include this in the 0.x branch, but it might be worthwhile to consider for a future [json-c-1.0](https://github.com/json-c/json-c/wiki/Proposed-changes-for-new-v1.0-version).

I'm not convinced the additional memory overhead is worth it.  What % impact does that have on total memory usage?
Also, you're mixing a change to storing the hash with a change to how the array index is calculated (using bitwise-& instead of %), so it's not clear what portion of the performance improvement comes from which change.

hawicz:
Closing, since this won't be merged as-is into the current branch.

PR #589, opened 2020-05-02
Detect broken RDRAND during initialization

### LINKED ISSUE #588: Broken RDRAND causes infinite looping
Some CPUs, like the Ryzen 3600X, report having `RDRAND` in `cpuid`, but actually executing `RDRAND` returns an endless stream of `0xFFFFFFFF`. This causes json-c to lock up retrying `RDRAND`, but never trying another source of entropy.

https://github.com/json-c/json-c/blob/31f1ab2be1142d982fc9b8c8913bfd1233599ebb/linkhash.c#L467-L469

A number of Arch Linux users reported issues with this in https://github.com/swaywm/sway/issues/5290, where json-c was recently upgraded from 0.13 to 0.14. There, this problem manifested as the window manager entirely locking up during JSON IPC.

The kernel supports a `nordrand` flag, but that has no effect on the availability of `RDRAND` signaled by `cpuid`. The workaround here is to update BIOS to a version that fixes this issue, but considering the number of reports in that project alone, it might be worthwhile to include fallback logic in json-c itself, and use a different random source if `RDRAND` fails too many times. I'm happy to submit such a patch, if desired.

Xyene:
Some CPUs advertise `RDRAND` in `CPUID`, but return `0xFFFFFFFF`
unconditionally. To avoid locking up later, test `RDRAND` during
initialization, and if it returns `0xFFFFFFFF`, mark it as nonexistent.

Fixes #588.

### DIFF
--- a/random_seed.c
+++ b/random_seed.c
@@ -43,12 +43,41 @@ static void do_cpuid(int regs[], int h)
 
 #if HAS_X86_CPUID
 
+static int get_rdrand_seed(void);
+
+// Valid values are -1 (haven't tested), 0 (no), and 1 (yes).
+static int _has_rdrand = -1;
+
 static int has_rdrand(void)
 {
-	// CPUID.01H:ECX.RDRAND[bit 30] == 1
-	int regs[4];
-	do_cpuid(regs, 1);
-	return (regs[2] & (1 << 30)) != 0;
+	if (_has_rdrand == -1)
+	{
+		// CPUID.01H:ECX.RDRAND[bit 30] == 1
+		int regs[4];
+		do_cpuid(regs, 1);
+		if (!(regs[2] & (1 << 30)))
+		{
+			_has_rdrand = 0;
+		} else
+		{
+			// Some CPUs advertise RDRAND in CPUID, but return 0xFFFFFFFF
+			// unconditionally. To avoid locking up later, test RDRAND here. If over
+			// 10 trials RDRAND has returned the same value, declare it broken.
+			_has_rdrand = 0;
+			int prev = get_rdrand_seed();
+			for (int i = 0; i < 10; i++) {
+				int temp = get_rdrand_seed();
+				if (temp != prev) {
+					_has_rdrand = 1;
+					break;
+				}
+
+				prev = temp;
+			}
+		}
+	}
+
+	return _has_rdrand;
 }
 
 #endif


TheAvidDev:
I participated in swaywm/sway#5290 and can confirm that this patch fixes the issue.

random_seed.c
```diff
@@ -54,12 +54,16 @@ static void do_cpuid(int regs[], int h)
...
+
 static int has_rdrand(void)
 {
 	// CPUID.01H:ECX.RDRAND[bit 30] == 1
 	int regs[4];
 	do_cpuid(regs, 1);
-	return (regs[2] & (1 << 30)) != 0;
+	// Some CPUs advertise RDRAND in CPUID, but return 0xFFFFFFFF unconditionally.
+	// To avoid locking up later, test RDRAND here.
+	return (regs[2] & (1 << 30)) != 0 && get_rdrand_seed() != -1;
```
hawicz, line 66: Let's:
a) only do the extra check for AMD Ryzen processors.  It looks like these are CPU family 0x17, which I believe we can get from regs[0] by adding the values at index 8-11 and index 20-27 together (https://en.wikipedia.org/wiki/List_of_AMD_CPU_microarchitectures, https://en.wikipedia.org/wiki/CPUID)
b) check several values (e.g. 10) and only if they're _all_ -1 disable rdrand
c) persist that in a flag variable, and don't bother even calling cpuid if we've already determined RDRAND is broken.

TheAvidDev: Is there any reason to limit this check to only AMD CPUs and not just all CPUs?

Hard coding a list of CPU families (which I believe is actually `0x15` and `0x16`) seems prone to potentially requiring future changes.

hawicz: Because it's a check specifically to workaround a bug in AMD CPUs, not a general-purpose "is this random source good" check.  The next broken CPU might returns all zeros, or any other value, or even values that vary but are actually non-random in some way, and we shouldn't give the impression that a simple "is it all -1's" check would do anything to catch those.

Where are you getting your CPU family values from ?  The wikipedia page I'm looking at lists 0x15 and 0x16 as the lines of processors _before_ the Ryzen ones.

TheAvidDev: Ah, it seems I *might* be incorrect about `0x15` and `0x16`. These processors *are* affected by a bug that means `RDRAND` is bad after suspending the system, as per this [lkml patch](https://lkml.org/lkml/2019/8/14/1374), but the kernel will automatically disable their `RDRAND` cpuid bit as seen [here](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpu/amd.c#L848)

However, it also looks like this disabling [can fail](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpu/amd.c#L872).

ploxiln: Here's another, slightly more generic, idea: take three RDRAND samples. If they are all the same, it is astronomically unlikely that RDRAND is not broken (about 1 chance in 2**64). (The chance of a working RDRAND returning -1 is about 1 chance in 2**32, aka 1 in 4 billion.)

Xyene: @ploxiln: that's what the kernel does, with 8 samples. systemd bails if the first value it sees is -1, and doesn't try again. (I did a bit of reading to see how other projects deal with this.)

I wouldn't mind adding the AMD CPU check, but I feel as if whatever false positives that'd prevent would be dwarfed by several orders of magnitude by just trying RDRAND another round or two. With 10 rounds, there's a 0.00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004% chance of misdetecting (4.68e-95%). The extra CPUID check would add an insignificant few more 0s, at the cost of more complexity.

hawicz: My point wasn't really to raise the bar to avoid false positives, but more as self-documenting code that makes it clear why the special check is there in the first place.  If there are other known issues with rdrand, then having this be a more general case check, and simply documenting in a comment what processors we had in mind when the code was written would be fine.
Thanks for reminding me the low false positive rate; given that, the approach of "take three RDRAND samples" and verify they're not all the same sounds good.

Xyene: Again it seems we raced in commenting / committing :)

I've changed the code around to be more robust in detecting not-randomness, if it sees the same value 10 times in a row it declares the random generator as broken. This presumably should be useful for anything that fails to return randomness that's not necessarily a Ryzen and not necessarily -1.

I can still add the CPUID check if you wish, my only concern there is that it's another moving part that might add complexity in the long run.

PR #590, opened 2020-05-02
Fix segmentation fault in CPUID check

Xyene:
The current CPUID check seems to be reliant on the state of registers at call-time (...or something like that?), since it occasionally segfaults. Here's an example of it in a simplified context:

```c
#include <stdio.h>

static void do_cpuid(int regs[], int h)
{
	/* clang-format off */
    __asm__ __volatile__(
#if defined __x86_64__
                         "pushq %%rbx;\n"
#else
                         "pushl %%ebx;\n"
#endif
                         "cpuid;\n"
#if defined __x86_64__
                         "popq %%rbx;\n"
#else
                         "popl %%ebx;\n"
#endif
                         : "=a"(regs[0]), [ebx] "=r"(regs[1]), "=c"(regs[2]), "=d"(regs[3])
                         : "a"(h));
	/* clang-format on */
}

static int has_rdrand(void)
{
    // CPUID.01H:ECX.RDRAND[bit 30] == 1
    int regs[4];
    do_cpuid(regs, 1);
    return (regs[2] & (1 << 30)) != 0;
}

static int get_rdrand_seed(void)
{
	int _eax;
	// rdrand eax
	/* clang-format off */
	__asm__ __volatile__("1: .byte 0x0F\n"
	                    "   .byte 0xC7\n"
	                    "   .byte 0xF0\n"
	                    "   jnc 1b;\n"
	                    : "=a" (_eax));
	/* clang-format on */
	return _eax;
}

int main(void) { 
	printf("HAS RDRAND: %d\n", has_rdrand());
	for (int i = 0; i < 10; i++) {
		printf("%d\n", get_rdrand_seed());
	}
	return 0;
}
```

Compiling and running, the result is:

```
$ gcc foo.c -g -o foo && ./foo
Segmentation fault
```

Poking in GDB,

```
Program received signal SIGSEGV, Segmentation fault.
0x0000555555555165 in do_cpuid (regs=0x0, h=1) at foo.c:6
6	   __asm__ __volatile__(
(gdb) disas 0x0000555555555150,0x0000555555555170
Dump of assembler code from 0x555555555150 to 0x555555555170:
   0x0000555555555150 <do_cpuid+27>:	mov    -0x8(%rbp),%rax
   0x0000555555555154 <do_cpuid+31>:	lea    0xc(%rax),%r8
   0x0000555555555158 <do_cpuid+35>:	mov    -0xc(%rbp),%eax
   0x000055555555515b <do_cpuid+38>:	push   %rbx
   0x000055555555515c <do_cpuid+39>:	cpuid  
   0x000055555555515e <do_cpuid+41>:	pop    %rbx
   0x000055555555515f <do_cpuid+42>:	mov    %eax,%edi
   0x0000555555555161 <do_cpuid+44>:	mov    -0x8(%rbp),%rax
=> 0x0000555555555165 <do_cpuid+48>:	mov    %edi,(%rax)
   0x0000555555555167 <do_cpuid+50>:	mov    %esi,(%r10)
   0x000055555555516a <do_cpuid+53>:	mov    %ecx,(%r9)
   0x000055555555516d <do_cpuid+56>:	mov    %edx,(%r8)
```

I'm not sure what the pushing in the existing code tries to accomplish, but replacing it with something like [what the Linux kernel](http://lxr.linux.no/#linux+v2.6.39/arch/x86/include/asm/processor.h#L173) and [this thread](https://gcc.gnu.org/legacy-ml/gcc-help/2015-08/msg00089.html) do resolves the issue.

### DIFF
--- a/random_seed.c
+++ b/random_seed.c
@@ -26,19 +26,8 @@
 static void do_cpuid(int regs[], int h)
 {
 	/* clang-format off */
-    __asm__ __volatile__(
-#if defined __x86_64__
-                         "pushq %%rbx;\n"
-#else
-                         "pushl %%ebx;\n"
-#endif
-                         "cpuid;\n"
-#if defined __x86_64__
-                         "popq %%rbx;\n"
-#else
-                         "popl %%ebx;\n"
-#endif
-                         : "=a"(regs[0]), [ebx] "=r"(regs[1]), "=c"(regs[2]), "=d"(regs[3])
+    __asm__ __volatile__("cpuid"
+                         : "=a"(regs[0]), "=b"(regs[1]), "=c"(regs[2]), "=d"(regs[3])
                          : "a"(h));
 	/* clang-format on */
 }


TheAvidDev:
As with #589 I had this issue and this patch fixed it.

hawicz:
The constraint flags on the output operands (`"=a"(regs[0])`, etc...) look like they should be equivalent, though I'm not quite sure why the existing code uses `[ebx] "=r"` instead of just "=b".  The latter syntax seems better, so let's keep that part of this change.
In any case, I'd expect that the compiler would recognize when ebx is used, and save/restore it itself in the surrounding function, so doing that inside the asm block is unnecessary.  Looking at the assembly code, that save/restore _does_ happen.
(see the answer which quotes MS docs at https://stackoverflow.com/questions/10504984/why-does-gcc-inline-assembler-require-clobbering-information-but-msvc-doesnt)
The other difference here, with  the `"c"(0)` input operand, just sets ECX to 0.  Per the Wikipedia page on CPUID the ECX register selects "sub-leaves" of the leaf selected by the EAX register value, but for EAX=0x01 that should be entirely irrelevant.
Please leave that `"c"(0)` off of this change.

On to what actually fixes the crashing problem:
rbx/ebx is defined as a calle-saved register, so I'd say that the existing code is trying to explicitly save and restore it to ensure that the cpuid instruction doesn't clobber it.  .  However, normally, I'd expect push/pop to be safe, but it looks like something goes wrong with the stack pointer handling.
The "args" parameter is passed in to do_cpuid() in the RDI register, then the function prologue writes it out to the stack, but *doesn't adjust RSP* !  So then when the asm code uses push/pop, it overwrites the value with whatever is in $ebx.
This feels like a bug in the compiler, but given that I think the push/pop is unnecessary, since we're not implementing a function that's being called, just a bit of code inside a function, dropping those instructions sounds good to me.

I have no idea how this ever works with the code as-is.

hawicz:
I did a bit more digging, and the reason I, personally am not running into this issue is because I rarely build with ENABLE_RDRAND=Yes, and where I usually work from is a machine that doesn't support the rdrand instruction.  oops.

When running within a program that has done a few more things than your sample program before calling json_c_get_random_seed()  (e.g. apps/json_parse), it seems we just get lucky, and:
* the $eax return value from cpuid is written into some arbitrary location that $rbx happened to be set to on entering do_cpuid() (probably some earlier malloc'd piece of memory)
* the $ebx, $ecx, $edx values from cpuid are written to the proper location because the memory addresses of `regs[1]`, `regs[2]` and `regs[3]` happen to be loaded into registers *before* the asm block.

Anyway, I'm going to merge this, and make one slight tweak afterwards, but I'd still like to figure out how use of push/pop or any instructions that affect $esp/$rsp is supposed to work.  (or perhaps it's _not_ supposed to work?)

Xyene:
I imagine not adjusting RSP might be an optimization on the part of the compiler; if there's nothing in the function that could mess with it, it might be deciding that it's not worth emitting those instructions. As far as I know, GCC treats the ASM as an opaque blob, so it doesn't know that it's touching the stack.

> I have no idea how this ever works with the code as-is.

Seems to depend on optimization level :)

At -O1, `do_cpuid` gets inlined, and problematic prologue/epilogue doesn't exist. The inlining threshold might be larger on the actual json-c.

hawicz:
Oh, you updated your code, so I don't need to adjust anything, thanks!  (\*grumble\* stale github page responses \*grumble\*)

hawicz:
I found an explanation about the hazards of stack operations with inline assembly over at https://stackoverflow.com/questions/48853757/gcc-inline-assembly-with-stack-operation
I believe the particularly problematic bit is the ["red zone"](https://stackoverflow.com/questions/38042188/where-exactly-is-the-red-zone-on-x86-64) 

I also noticed that that page points at https://stackoverflow.com/questions/14266772/how-do-i-call-cpuid-in-linux, which points out that GCC provides a cpuid.h header and a `__get_cpuid()` function, so json-c's `do_cpuid()` function could be replaced with that.  A task for another day though; I've got other things to work on.

PR #591, opened 2020-05-02
Update README.md

stackerzzq:


hawicz:
Not merging, cmake-configure is supposed to accept the `--prefix=<foo>` syntax, and I just fixed it to do so (commit f9605e9).

Issue #593, opened 2020-05-04
Building both static and shared libraries


jsmolic:
Hi,
I'm currently using this package on Gentoo Linux, and I have a question on building static libraries. I would like to know whether it would be possible to build both static and shared library. I've investigated a bit, and if I'm not mistaken `CMakeLists` only enables either static or shared library, not both. I would like to add config option in Gentoo package which would enable building both static and shared libraries if static-libs support is selected. Would it be possible to do this? Thank you for answer.

Kind regards,
Jakov

hawicz:
d'oh!  I knew there was something I forgot to get in to the last release.
I hope it's possible to do, but I don't know how off hand.  If you can dig up how to get cmake to do this, or even provide a PR, it would be greatly appreciated.

jsmolic:
It should be doable, I'll investigate and try to come up with something when I catch some time :)

jsmolic:
@dota17 @hofnarrr Thank for fixing this

PR #595, opened 2020-05-06
Support to build both static and shared libraries

dota17:
#593

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -403,14 +403,14 @@ target_include_directories(${PROJECT_NAME}
 
 # Allow to build static and shared libraries at the same time
 if (BUILD_STATIC_LIBS)
-    set(ORIGINAL_STATIC_LIB_NAME ${PROJECT_NAME}-static)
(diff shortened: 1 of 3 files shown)


dota17:
Allow to build both static and shared libraries.

```sh
cmake -DBUILD_STATIC_LIBS=ON ..
```

hawicz:
We should default to both static and shared being built, and describe `-DBUILD_SHARED_LIBS=OFF` and `-DBUILD_STATIC_LIBS=OFF` as simply turning the relevant build off, rather than having it say it turns the "other" libs build on.

dota17:
ok

hofnarrr:
Hi, there is some problems with the fix this branch provides. Not sure what's the "correct" way to submit fixes to other people's PR's, so I submitted a pull request to dota17's branch (https://github.com/dota17/json-c/pull/90).

dota17:
@hofnarrr  Thanks a lot! I found the install error of share&static libraries  before and was thinking how to fix it. ha-ha.

PR #602, opened 2020-05-10
fix json_parse_uint64() usage of errno

### LINKED ISSUE #601: json_tokener_parse() in master sets errno=1 "Operation not permitted"
json_tokener_parse() is setting errno=1 on my system with current git master, even when successful.

It did not do this with json-c v0.12.

Unfortunately I'm not  sure how to trace it further.

ploxiln:
introduced in #542
fixes #601

### DIFF
--- a/json_util.c
+++ b/json_util.c
@@ -245,19 +245,17 @@ int json_parse_uint64(const char *buf, uint64_t *retval)
 {
 	char *end = NULL;
 	uint64_t val;
-	errno = 1;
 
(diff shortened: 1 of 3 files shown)


hawicz:
er.. this is a bit more than just "fixing errno" usage.   Really, you're fixing the wonky "did it work" logic in json_parse_uint64(), and optimizing that function to skip the strtoull calls for negatives, which is worthwhile so I'll merge it.

PR #603, opened 2020-05-10
Backport of fixes from master branch.

besser82:
This pull-request backports all fixes that have been applied on the master branch to json-c-0.14.  It also adds the option to build a static lib along with the shared library.


hawicz:
It seems to me that backporting *all* the fixes is a bit silly.  Rather than doing that it'd be better to simply create a new 0.15 release.

besser82:
> It seems to me that backporting _all_ the fixes is a bit silly. Rather than doing that it'd be better to simply create a new 0.15 release.

If you would do a 0.15 release in the near future then, that would be really nice.

PR #606, opened 2020-05-13
Improved support for IBM operating systems

davidjmccann:
Fix compiler errors and warnings when building on IBM operating systems such as AIX and IBM i.


json_object.c
```diff
@@ -1478,7 +1485,7 @@ static int json_object_copy_serializer_data(struct json_object *src, struct json
 	if (dst->_to_json_string == json_object_userdata_to_json_string ||
 	    dst->_to_json_string == _json_object_userdata_to_json_string)
 	{
-		dst->_userdata = strdup(src->_userdata);
+		dst->_userdata = strdup((char*)src->_userdata);
```
hawicz, line 1488: Is this actually needed?  I really don't want to be littering this code with casts to cater to a broken compiler if we can avoid it.  :(

tests/test_parse.c
```diff
@@ -446,7 +446,7 @@ static void test_incremental_parse()
 		json_tokener_set_flags(tok, step->tok_flags);
 
 		if (length == -1)
-			length = strlen(step->string_to_parse);
+			length = (int)strlen(step->string_to_parse);
```
hawicz, line 449: How about changing `int length` to be size_t instead?

json_object.c
```diff
@@ -901,18 +903,23 @@ static int json_object_double_to_json_string_format(struct json_object *jso, str
...
+#ifdef HAVE_DECL_ISNAN
 	if (isnan(jso->o.c_double))
 	{
 		size = snprintf(buf, sizeof(buf), "NaN");
 	}
-	else if (isinf(jso->o.c_double))
+	else
+#endif
+#ifdef HAVE_DECL_ISINF
+	if (isinf(jso->o.c_double))
```
hawicz, line 914: eh?  These systems have `INFINITY` (as referenced in json_tokener.c), but no way to check if a number is that value?

tests/test_parse.c
```diff
@@ -149,10 +149,10 @@ static void test_utf8_parse()
 	// json_tokener_parse doesn't support checking for byte order marks.
 	// It's the responsibility of the caller to detect and skip a BOM.
 	// Both of these checks return null.
-	char utf8_bom[] = {0xEF, 0xBB, 0xBF, 0x00};
-	char utf8_bom_and_chars[] = {0xEF, 0xBB, 0xBF, '{', '}', 0x00};
-	single_basic_parse(utf8_bom, 0);
-	single_basic_parse(utf8_bom_and_chars, 0);
+	unsigned char utf8_bom[] = {0xEF, 0xBB, 0xBF, 0x00};
```
hawicz, line 152: Instead of changing the type and casting here, would it work to use e.g. "\xef\xbb\xbg" instead?

json_object.c
davidjmccann: Correct. On IBM i the printf function will automatically turn a value of INFINITY to "infinity", and NAN to "nan". There are no isinf or isnan functions.

ploxiln: for what it's worth, the python json decoder will accept `{"val": Infinity}` and `{"val": NaN}`, but will not accept `{"val": infinity}` or `{"val": nan}` (and ECMA 262 dictates the capitalized versions)

hawicz: hmm... that's not going to work so well with the tests, since they're going to be looking for "Infinity" and "NaN".
Are you sure they aren't available if you include float.h?
Or, perhaps _XOPEN_SOURCE needs to be defined? (e.g. like mentioned at https://github.com/matplotlib/matplotlib/issues/12535/)

hawicz: If they're not provided, we'll need to provide our own.  Can you please test this program and see if it works?:
```
#include <stdlib.h>
#include <stdio.h>
#include <float.h>

int main()
{
    double f[] = {
        0.0d, 1.0d, DBL_MAX, 0.0d / 0.0d, 1.0d/0.0d, -1.0d/0.0d
    };
    int i;

    for (i = 0; i < 6; i++) {
        double x = f[i];

        int is_nan = (x != x);
        int is_inf = (x < -DBL_MAX || x > DBL_MAX);

        printf("%20g%4d%4d\n", x, is_nan, is_inf);
    }

    return 0;
}
```

It should produce output like:
```
                   0   0   0
                   1   0   0
        1.79769e+308   0   0
                -nan   1   0
                 inf   0   1
                -inf   0   1
```

davidjmccann: There's definitely no isnan or isinf on IBM i according to the Knowledge Center (https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_74/rtref/rtrefmain.htm). I will run your test program on IBM i and let you know the results when they are available.

tests/test_parse.c
davidjmccann: That doesn't really make sense, given that size_t is unsigned and the value -1 is expected.

json_object.c
davidjmccann: No, you're right, I've managed to get around this one.

tests/test_parse.c
davidjmccann: I'll give it a go.

tests/test_parse.c
hawicz: Oh, right.  never mind then, let's leave it with the cast to int in place then.

json_object.c
davidjmccann: Your original program crashed. I modified it to look like this:
const unsigned int lnIND[2] = {0x00000000, 0xfff80000};
const unsigned int lnINF[2] = {0x00000000, 0x7ff00000};

int main()
{
  const double ZERO                 = 0.0;
  const double ONE                  = 1.0;
  const double AN_INDETERMINATE     = *( double* )lnIND;
  const double AN_INFINITY_POSITIVE = *( double* )lnINF;
  const double AN_INFINITY_NEGATIVE = -AN_INFINITY_POSITIVE;

  double f[] = {ZERO, ONE, DBL_MAX, AN_INDETERMINATE, AN_INFINITY_POSITIVE, AN_INFINITY_NEGATIVE};

It produced this output:
```
                   0   0   0
                   1   0   0
        1.79769e+308   0   0
        2.12174e-314   0   0
        1.06048e-314   0   0
       -1.06048e-314   0   0
```

davidjmccann: I'm not sure how else to get it to produce a Nan or a value for Infinity.

hawicz: Crashes how?  Do the machines you're on not properly handle floating point divide-by-zero traps?  Is it defined to use something other than the IEEE 754 standard for floating point math?

hawicz: Also, it seems like you haven't managed to actually get real NaN and Infinity values in the test program, since you said that printf would emit "nan" or "infinity", but that isn't happening.
How about trying to use the `INFINITY` and `NAN` constants?  You might need to include math.h to get those.

davidjmccann: Using INFINITY and NAN I get this:
```
                    0   0   0  
                    1   0   0  
         1.79769e+308   0   0  
                  nan   1   0  
             infinity   0   1  
            -infinity   0   1  
```

json_object.c
davidjmccann: Now removed from the changes.

tests/test_parse.c
davidjmccann: Now changed to using a string as suggested.

json_object.c
hawicz: Perfect!  I'll merge this PR as-is, then fix things up afterwards.

hawicz:
@davidjmccann, can you please check that it still works for with with the most recent change in commit 1526c84?

davidjmccann:
@hawicz Yes, that still builds fine with your recent changes. Thanks for doing that.

PR #610, opened 2020-05-15
let's not call lh_table_resize with INT_MAX

ddstreet:
That would be bad.

### DIFF
--- a/linkhash.c
+++ b/linkhash.c
@@ -584,7 +584,7 @@ int lh_table_insert_w_hash(struct lh_table *t, const void *k, const void *v, con
 	{
 		/* Avoid signed integer overflow with large tables. */
 		int new_size = (t->size > INT_MAX / 2) ? INT_MAX : (t->size * 2);
-		if (t->size == INT_MAX || lh_table_resize(t, new_size) != 0)
+		if (new_size == INT_MAX || lh_table_resize(t, new_size) != 0)
 			return -1;
 	}
 


besser82:
> That would be bad.

Why?  `INT_MAX` is a perfectly valid size…

linkhash.c
```diff
@@ -584,7 +584,7 @@ int lh_table_insert_w_hash(struct lh_table *t, const void *k, const void *v, con
 	{
 		/* Avoid signed integer overflow with large tables. */
 		int new_size = (t->size > INT_MAX / 2) ? INT_MAX : (t->size * 2);
-		if (t->size == INT_MAX || lh_table_resize(t, new_size) != 0)
+		if (new_size == INT_MAX || lh_table_resize(t, new_size) != 0)
```
besser82, line 587: Idon't get the purpose here.  Why shouldn't we resize to `INT_MAX`?  The condotional was fine before, as we do not resize the table above `INT_MAX`.  Anyways, `INT_MAX` is a petrfectly valid size.

ddstreet: setting new_size but checking t->size is actually what you wanted to do?

besser82: The actual code ensures the following:

  1. `lh_table_resize` will be called to resize the table, if it needs to be resized.  `new_size` is computed to be the double of the current size of the table, but *not* more than `INT_MAX`.
  2. If the table is already `INT_MAX` of size, there will be no resizing operation, nor any insertion. Instead the return value of the function will indicate an error.

ddstreet: i see your logic.  IMHO 1) that's extremely bad code because it looks a *lot* like a set a, check b error; and 2) is doubling the size from 1G -> 2G a good idea?

In any case, if it's correct, then ignore my PR :)

FYI, I'm coming here due to https://bugs.launchpad.net/ubuntu/+source/json-c/+bug/1878723

ddstreet: (to clarify, I assumed you were capping the table size at INT_MAX / 2, and using INT_MAX as an error marker to avoid any more resizing)

besser82: > i see your logic. IMHO 1) that's extremely bad code because it looks a _lot_ like a set a, check b error; and 2) is doubling the size from 1G -> 2G a good idea?

I don't think it will pratically affect any real world scenario, as having JSON data with 1+ Gi count of nested and/or single objects is quite unlikely.
 
> FYI, I'm coming here due to https://bugs.launchpad.net/ubuntu/+source/json-c/+bug/1878723

FYI, The correct fix for that bug has already been mentioned there: https://bugs.launchpad.net/ubuntu/+source/json-c/+bug/1878723/comments/3

ddstreet: I don't see how that's any change; < is higher in order of operations than ternary.

besser82: Yes, but the conditional actually was backwards.

`int new_size = INT_MAX / 2 < t->size ? t->size * 2 : INT_MAX;`

If `INT_MAX / 2` is smaller than `t->size`, it would resize to double `t->size`, otherwise resize to `INT_MAX`.

The fix changed it to have the correct behaviour:

`int new_size = (t->size > INT_MAX / 2) ? INT_MAX : (t->size * 2);`

If `INT_MAX / 2` is smaller than `t->size`, resize to `INT_MAX`, otherwise resize to double `t->size`.

***

Check this example:

```
#include <stdio.h>

int main (void)
{
  for (int i = 0; i <=3; ++i)
    {
      printf ("wrong: 4 / 2 < %d ? 0 : 1 == %d\n", i, 4 / 2 < i ? 0 : 1);
      printf ("right: (%d > 4 / 2) ? 1 : 0 == %d\n", i, (i > 4 / 2) ? 1 : 0);
    }
}
```

stdout:

```
wrong: 4 / 2 < 0 ? 0 : 1 == 1
right: (0 > 4 / 2) ? 1 : 0 == 0
wrong: 4 / 2 < 1 ? 0 : 1 == 1
right: (1 > 4 / 2) ? 1 : 0 == 0
wrong: 4 / 2 < 2 ? 0 : 1 == 1
right: (2 > 4 / 2) ? 1 : 0 == 0
wrong: 4 / 2 < 3 ? 0 : 1 == 0
right: (3 > 4 / 2) ? 1 : 0 == 1
```

ddstreet: > If INT_MAX / 2 is smaller than t->size, it would resize to double t->size, otherwise resize to INT_MAX.

ah.  well if you checked for new_size == INT_MAX it would have caught that error xD

anyway, personally i think the logic is too clever and will cause more trouble later (code is about readability, sometimes that's more important than being concise and clever).  But yeah, I see what you're doing now.

PR #625, opened 2020-05-25
json-c-0.13.x: Detect broken RDRAND during initialization.

besser82:
Some CPUs advertise RDRAND in CPUID, but return 0xFFFFFFFF unconditionally.  To avoid locking up later, test RDRAND during initialization, and if it returns 0xFFFFFFFF, mark it as nonexistent.

Also fix a possible segmentation fault in CPUID check.

This commit is a squashed and slightly modified backport of the following commits on the master branch:

  * 0e5bbcaa162ac7850eb4fcd8f91391837d0efb50
  * 4d36b0287d3ab0912ba8a4790340ca099960b2b0
  * 80863140263be5f2dc630938ed8f0066f8a1ab43

### DIFF
--- a/random_seed.c
+++ b/random_seed.c
@@ -26,20 +26,11 @@
 
 static void do_cpuid(int regs[], int h)
 {
-    __asm__ __volatile__(
-#if defined __x86_64__
-                         "pushq %%rbx;\n"
-#else
-                         "pushl %%ebx;\n"
-#endif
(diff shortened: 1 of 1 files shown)


hawicz:
As with PR #612, I don't want to give the impression that old branches are truly "active", but since there are likely active users I'm merging this.

PR #626, opened 2020-05-28
Fixes for cmake 2.8.12 + link issue on AIX 6.1/cc 11.01

pointbre:
# Fixes for cmake 2.8.12

## Build failed with cmake 2.8.12
```
$ cmake -DCMAKE_INSTALL_PREFIX=/home/xxx/json-c ./json-c-json-c-0.13.1-20180305
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_VERSION_COMPILER_ENV_VAR
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_VERSION_COMPILER
CMake Error: Could not find cmake module file: /home/xxx/json-c/CMakeFiles/2.8.12.2/CMakeVERSIONCompiler.cmake
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_0.13.1_COMPILER_ENV_VAR
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_0.13.1_COMPILER
CMake Error: Could not find cmake module file: /home/xxx/json-c/CMakeFiles/2.8.12.2/CMake0.13.1Compiler.cmake
CMake Error: Could not find cmake module file: CMakeVERSIONInformation.cmake
CMake Error: Could not find cmake module file: CMake0.13.1Information.cmake
CMake Error: CMAKE_VERSION_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_0.13.1_COMPILER not set, after EnableLanguage
/home/xxx/json-c/json-c-json-c-0.13.1-20180305
```
## With the fixes
```
$ cmake -DCMAKE_INSTALL_PREFIX=/home/xxx/json-c/lib ../json-c
/home/xxx/json-c/json-c
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal
configure.ac:1: error: Autoconf version 2.64 or higher is required
configure.ac:1: the top level
autom4te: /usr/bin/m4 failed with exit status: 63
aclocal: autom4te failed with exit status: 63
autoreconf: aclocal failed with exit status: 63
checking for a BSD-compatible install... /usr/bin/install -c
...
-- Configuring done
-- Generating done
-- Build files have been written to: /home/xxx/json-c/build

$ make
[ 50%] Built target json-c
[100%] Built target json-c-static

$ make install
[ 50%] Built target json-c
[100%] Built target json-c-static
Install the project...
-- Install configuration: ""
-- Installing: /home/xxx/json-c/lib/lib/libjson-c.so
-- Installing: /home/xxx/json-c/lib/lib/libjson-c.a
...
```
## Test with hello world application
```
gcc -o test.out test.c -L/home/xxx/json-c/json-c-0.13.1-built/lib -ljson-c -I/home/xxx/json-c/json-c-0.13.1-built/include/json-c

$ LD_LIBRARY_PATH=/home/xxx/json-c/json-c-0.13.1-built/lib ./test.out
JSON string: {"sitename" : "joys of programming" }
type: type: json_type_stringn          value: joys of programmingn

JSON-c version: 0.13.1
```

## Test environment
* Cent OS 6.8
* Cmake 2.8.12

# Linking issue on AIX 6.1 with cc V11.1
When linking applicatio to json-c-0.13.1(installed by json-c-0.13.1-1.aix5.1.ppc.rpm from http://www.oss4aix.org/download/RPMS/json-c/), cc complained heaps of things which were fixed by the change.

## Test environment
* AIX 6.1
* cc -qversion: IBM XL C/C++ for AIX, V11.1 (5724-X13), Version: 11.01.0000.0021
* json-c-0.13.1-1.aix5.1.ppc.rpm from http://www.oss4aix.org/download/RPMS/json-c/


CMakeLists.txt
```diff
@@ -1,9 +1,28 @@
...
+set(LIB_MAJOR_VERSION "0")
+set(LIB_MINOR_VERSION "13")
+set(LIB_PATCH_VERSION "1")
+set(LIB_VERSION_STRING "${LIB_MAJOR_VERSION}.${LIB_MINOR_VERSION}.${LIB_PATCH_VERSION}")
+
+if(POLICY CMP0048)
+  cmake_policy(SET CMP0048 NEW)
+endif()
+
+set(LIB_MAJOR_VERSION "0")
```
hawicz, line 16: Why are you setting these variables twice?

hawicz: Actually, why set these at all, if we're not using them anywhere?  It seems like there's something missing here.

pointbre: 1.Sorry for that. I will update it again. 2.cmake 2.8 doesn't support VERSION in project, so instead of project(json-c VERSION 0.13.1), LIB_VERSION_STRING needs to be used.

linkhash.h
```diff
@@ -348,7 +348,10 @@ int lh_table_resize(struct lh_table *t, int new_size);
 /**
  * @deprecated Don't use this outside of linkhash.h:
  */
-#if !defined(_MSC_VER) || (_MSC_VER > 1800)
+/* on AIX, cc doesn't like this */
+#if AIX_CC
```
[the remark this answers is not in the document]
pointbre: Again, mistake. I should have used `#ifdef AIX_CC`.

Would the following be ok?

```
#if (!defined(_MSC_VER) && !defined(AIX_CC)) || (_MSC_VER > 1800)
/* VS2010 can't handle inline funcs, so skip it there */
#define _LH_INLINE inline
#else
#define _LH_INLINE
#endif
```

CMakeLists.txt
hawicz: LIB_VERSION_STRING _isn't_ used.  You're setting all those variables for no apparent purpose.
However, we _do_ use the PROJECT_VERSION* variables, so I don't quite see how these changes could work.  I think you're going to need to set _those_ variables inside the < 3.0 block

pointbre: I've moved them as you suggested.

hawicz: I still don't understand how this can work.  What sets the PROJECT_VERSION* variables? 
Please try `make distcheck` to see if that is ok.

pointbre: What I know for sure is that cmake 2.8.12 doesn't support version with project command. The below is from https://cmake.org/cmake/help/v2.8.12/cmake.html#command:project.

```
project: Set a name for the entire project.
  project(<projectname> [languageName1 languageName2 ... ] )
Sets the name of the project. Additionally this sets the variables <projectName>_BINARY_DIR and <projectName>_SOURCE_DIR to the respective values.

Optionally you can specify which languages your project supports. Example languages are CXX (i.e. C++), C, Fortran, etc. By default C and CXX are enabled. E.g. if you do not have a C++ compiler, you can disable the check for it by explicitely listing the languages you want to support, e.g. C. By using the special language "NONE" all checks for any language can be disabled.
```

Whereas, cmake 3.0.2 supports version in project command:

```
project(<PROJECT-NAME> [LANGUAGES] [<language-name>...])
project(<PROJECT-NAME>
        [VERSION <major>[.<minor>[.<patch>[.<tweak>]]]]
        [LANGUAGES <language-name>...])
```

So, I've made another change:

1.I have taken out version settings:
```
if (CMAKE_VERSION VERSION_LESS 3.0)
  project(json-c)
else()
  project(json-c VERSION 0.13.1)
endif()
```

2.Put version setting for cmake 2.8 and specify target VERSION properties
```
set_property(TARGET json-c PROPERTY C_STANDARD 99)
set_property(TARGET json-c-static PROPERTY C_STANDARD 99)
set_target_properties(json-c-static PROPERTIES OUTPUT_NAME json-c)

if (CMAKE_VERSION VERSION_LESS 3.0)
  # Fixes for CMake 2.8.12
  # Reference: https://github.com/Tencent/rapidjson/issues/1154
  set(LIB_MAJOR_VERSION "0")
  set(LIB_MINOR_VERSION "13")
  set(LIB_PATCH_VERSION "1")
  set(LIB_VERSION_STRING "${LIB_MAJOR_VERSION}.${LIB_MINOR_VERSION}.${LIB_PATCH_VERSION}")

  set_target_properties(json-c PROPERTIES VERSION ${LIB_VERSION_STRING})
  set_target_properties(json-c-static PROPERTIES VERSION ${LIB_VERSION_STRING})
endif ()
```

Here's the install result and I can see 0.13.1 is used(previously libjson-c.so.0.13.1 was not created):

```
Install the project...
-- Install configuration: ""
-- Installing: /prj/lucas/scm298514_9.32dev/extern/Linux/local64/lib/libjson-c.so.0.13.1
-- Installing: /prj/lucas/scm298514_9.32dev/extern/Linux/local64/lib/libjson-c.so
-- Installing: /prj/lucas/scm298514_9.32dev/extern/Linux/local64/lib/libjson-c.a
```
I wish this would make sense to you this time.

P.S. `make distcheck` is not available on 0.13.1, so I couldn't run it.

hawicz: Sorry, I didn't realize this was specific to the 0.13 branch.  Is there any reason you can't use 0.14, or even the current code?  I would rather not make fixes to a branch that haven't yet been done on master.

Also, even on the 0.13 branch there is a reference to PROJECT_VERSION.  Why are you refusing to use PROJECT_ VERSION* variables?!  Are you saying that cmake 2.8 will error out if you try to set any of those variables?

From my testing, if you _don't_ set those variables (well, at least PROJECT_VERSION) then the json-c.pc file ends up invalid, with an empty version.  (See the ` set(VERSION ${PROJECT_VERSION})` line)

pointbre: 1.I have to use same version on Cent OS 6.8 and AIX 6.1. there's are only json-c 0.12, 0.13, and 0.13.1, so 0.14 rpm is NOT available on AIX 6.1 at the moment, so I've tried to build 0.13.1 on Cent OS 6.8 which only has cmake 2.8.12 at the moment.

2.As I wrote, cmake 2.8.12 doesn't seem to support PROJECT_VERSION. I looked at the document of cmake 2.8.12 and there's nothing like PROJECT_VERSION. If you want, I can try later, but I'd like to ask you to have a look at document of cmake 2.8.12. Installing cmake 3.x would be another time consuming job which should be done by my company's IT team, so I wouldn't want to go that way so long as I can use same version of json-c.

3.With my latest change, as you can see,  no matter how it worked, `make install` installed `-- Installing: /prj/lucas/scm298514_9.32dev/extern/Linux/local64/lib/libjson-c.so.0.13.1`, so it should be ok.

hawicz: 1. ah, ok.  Though if 0.14 or the master branch don't build on AIX, I'd certainly like to know.  I adjusted CMakeLists.txt on master, and applied your fix to linkhash.h, so I expect that build to be ok if you care to go down that path.

2. It's irrelevant whether cmake supports (i.e. sets) that variable.  json-c's CMakeLists.txt **uses** it, so it needs to be set one way or another.  Please see commit fe308b8 for the fix that I pushed to master.
(incidentally, building cmake and installing locally isn't all that hard, though I can understand wanting to avoid the need to do that)

3. I'm not going to merge something that merely goes as far as not completely failing on `make install` when I know that it's generating at least _some_ bad output.  I cherry picked the version of the linkhash.h fix that I just pushed to master (1c6086a), but there have been too many changes to simply cherry pick the CMakeLists.txt change.  If you respin your PR so it sets the PROJECT_VERSION_* vars similar to fe308b8, I'll be happy to merge it.

pointbre: 1. I'd like to, but I won't be able to build 0.14 on AIX 6.1 as I am not allowed to install it on my own.

2 & 3. I will try the commit to see if it works and let you know tomorrow.

hawicz:
Closing without merging.  I pulled over the relevant pieces of fe308b8 to the json-c-0.13 branch, so this PR is no longer relevant.

PR #627, opened 2020-06-01
Compat fixes

sherpya:
there is no need to call CryptAcquireContextW because two string parameters are NULL
CryptAcquireContextW does not work on Win98, and CRYPT_SILENT flags is not supported on WInNT4 and Win98x, any info about WIn2k?

I'm returning the fallback function, if you don't like it it can be reverted, but anyway is a bad pratice to call exit() inside a library

### DIFF
--- a/random_seed.c
+++ b/random_seed.c
@@ -194,27 +194,35 @@ static int get_dev_random_seed(void)
 #pragma comment(lib, "advapi32.lib")
 #endif
 
+static int get_time_seed(void);
+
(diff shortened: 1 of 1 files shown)


sherpya:
do you prefer also against master?

hawicz:
Yes, if you could re-spin this against master that would be preferred, since this isn't a 0.14-specific change.
Also, I'm expecting to release a 0.15 version before a 0.14.1 (if at all), so it might not be worth pulling over to the 0.14 branch at all.

PR #628, opened 2020-06-03
get_cryptgenrandom_seed: compat with old windows + fallback

sherpya:
there is no need to call CryptAcquireContextW because two string parameters are NULL
CryptAcquireContextW does not work on Win98, and CRYPT_SILENT flags is not supported on WInNT4 and Win98x, any info about WIn2k?

I'm returning the fallback function, if you don't like it it can be reverted, but anyway is a bad pratice to call exit() inside a library

### DIFF
--- a/random_seed.c
+++ b/random_seed.c
@@ -226,27 +226,35 @@ static int get_dev_random_seed(void)
 #pragma comment(lib, "advapi32.lib")
 #endif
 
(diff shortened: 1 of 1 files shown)


sherpya:
I confirm it works also on win2k sp4 (with the flag enabled)

PR #632, opened 2020-06-07
Json object split

hawicz:
https://github.com/json-c/json-c/wiki/Proposal:-struct-json_object-split

An internal-only change (no public API differences) to split struct json_object into a base type (still called struct json_object) and several sub-types:

- struct json_object_object
- struct json_object_array
- struct json_object_boolean
- struct json_object_double
- struct json_object_int
- struct json_object_string

Changes are largely confined to json_object.c and json_object_private.h
As mentioned in the ChangeLog, this improves memory usage and processing speed.


dota17:
Sorry for the delay in the review.

json_object_private.h
```diff
@@ -43,36 +45,53 @@ struct json_object
 	json_object_private_delete_fn *_delete;
 	json_object_to_json_string_fn *_to_json_string;
```
dota17, line 46: Is it possible to extract these functions into public interface for further reduce the size of data structure? Three func pointers will occupy 24 bytes in a 64-bit system. This is also very effective.

json_object.c
```diff
@@ -1125,71 +1232,65 @@ int json_object_set_double(struct json_object *jso, double new_value)
...
+	if (len > (SSIZE_T_MAX - (sizeof(*jso) - sizeof(jso->c_string)) - 1))
+		return NULL;
+	objsize = (sizeof(*jso) - sizeof(jso->c_string)) + len + 1;
+	if (len < sizeof(void *))
+		// We need a minimum size to support json_object_set_string() mutability
+		// so we can stuff a pointer into pdata :(
+		objsize += sizeof(void *) - len;
+
+    jso = (struct json_object_string *)json_object_new(json_type_string, objsize,
+		&json_object_string_to_json_string, &json_object_string_delete);
```
dota17, line 1276: This will reduce one malloc call. Good!

dota17, line 1276: But I am curious about how much parformance will be improved if the `LEN_DIRECT_STRING_DATA` 32 is set to 8-12 according to the previous logic.

json_object.c
```diff
@@ -1202,49 +1303,75 @@ const char *json_object_get_string(struct json_object *jso)
...
+		// full size_t range.
+		return 0;
+
+	dstbuf = get_string_component_mutable(jso);
+	curlen = JC_STRING(jso)->len;
+	if (curlen < 0)
+		curlen = -curlen;
+	newlen = len;
+
+	if ((ssize_t)len > curlen)
```
dota17, line 1339: Due to the previous memory allocation, The logic here becomes more complicated. But considering that this function is used less, I think it's acceptable.

json_object_private.h
```diff
@@ -43,36 +45,53 @@ struct json_object
...
-				 */
-				char *ptr;
-				char data[LEN_DIRECT_STRING_DATA];
-			} str;
-			int len;
-		} c_string;
-	} o;
 	json_object_delete_fn *_user_delete;
 	void *_userdata;
+	char data[1]; // Actually the rest of a struct json_object_${o_type}
```
dota17, line 50: Is there any further plan here. It looks like a redundant data.

json_object.c
```diff
@@ -111,15 +195,23 @@ static void json_object_fini(void)
 
 /* helper for accessing the optimized string data component in json_object
  */
-static const char *get_string_component(const struct json_object *jso)
+static inline char *get_string_component_mutable(struct json_object *jso)
+{
+	if (JC_STRING_C(jso)->len < 0)
+	{
+		/* Due to json_object_str_string(), we might have a pointer */
```
dota17, line 202: modify:   str--->set   /* Due to json_object_set_string(), we might have a pointer */

json_object_private.h
hawicz: Are you suggesting, e.g., changing json_object_to_json_string_ext() to take a json_object_to_json_string_fn argument?  That won't work for two reason:
1. it changes the public API, and I definitely don't want to do that as part of _this_ change.
2. More importantly, it won't work, because each json_object is allowed to have a _different_ to_json_string function, specified by callers of the json-c API.

On the other hand, _delete can't be changed by external callers, so *that* we could potentially turn into direct calls in json_object_put.  Good idea, I'll take a stab at doing that.

json_object.c
hawicz: LEN_DIRECT_STRING_DATA is gone with these changes, so there in nothing to tune now.  In effect, *all* strings are direct now, with the amount allocated being the bare minimum needed.
If you want to check the performance of the code from master with LEN_DIRECT_STRING_DATA=8, vs. json_object-split code, you should be able to do that by making that change locally, then running something like `jc-bench.sh --before . --after json_object-split`

json_object_private.h
hawicz: You're right, I'll drop this field.

json_object.c
hawicz: 👍   Will fix.

json_object.c
hawicz: > 
> 
> Sorry for the delay in the review.

No worries!  Thanks for taking a look.

json_object.c
hawicz: d'oh, jc-bench.sh isn't quite working.  hmm... I'll need to fix something with it.

json_object_private.h
hawicz: See commit 5ebfeaed

json_object_private.h
hawicz: See commit 4c107121

json_object.c
hawicz: See commit 4c107121

json_object.c
hawicz: Actually jc-bench.sh is fine, just need to tweak how to run it a bit: `./jc-bench.sh --before origin/json_object-split`
With LEN_DIRECT_STRING_DATA=8 on the master branch, the decrease in memory usage is not quite as much (only 9% decrease vs. 12% decrease), which is expected.
The runtime improvement drops from 5.8% better to only 0.5% better.  I'm not sure why using 8 vs 32 makes the performance of master that much better all by itself, but even so the json_object-split branch still comes out (barely) ahead.

dota17: ok, I get it

Issue #621, opened 2020-05-18
json-c and jansson libraries have symbol conflicts


vysakh-av:
**Describe the bug**
In the application, we are using a shared library which is using json-c for its internal purpose. Later our application required to use a third party library, but its internally used jansson inside it for its functionality. Unfortunately when we run the application , it leads to memory corruption and process crashed.

Both of these libraries have multiple APIs with same name but different signature.
Due to this, there is a memory corruption that resulted in crash.

The following APIs are available in both json-c and jansson libraries with the same name:
json_object_get     
json_object_iter_next

**Steps To Reproduce**
This issue can easily reproduce with the following way:

- `Application` linked to libjson_c_lib.so and libjansson_c_lib.so. Both of these are wrapper libraries are using json-c and jannson internally.

- libjson_c_so wrapper library linked to libjson-c.so
- libjansson_c.so wrapper library linked to libjansson.so.

**Tried statically linked one of the library but its not helped here. If we statically link jansson to wrapper library, json-c calls failed especially json_tokener_parse(). If we statically link json-c library, its resulted in same memory corruption.**

hawicz:
So what do you want us to do?

vysakh-av:
Thank you for the support.
Here do we have any way to resolve this issue, we have tried different ways by statically linking and using -Bsymbolic linker flags, but it didn't help us. Any  help or support much appreciated.

micahsnyder:
This issue ties in with my request to build with `-fPIC` by default, so other libraries can link with the static libjson-c.a library.  Without `-fPIC`, you can only statically link with an application, meaning you can't rely on json-c as a dependency for another library because a downstream application may choose jansson and then crash with symbol collisions.

See my ticket for more details https://github.com/json-c/json-c/issues/508

smcv:
The possible ways to resolve this conflict go something like this:

* json-c renames its symbols to something like `jsonc_whatever()`, which is an API and ABI break, and presumably the json-c maintainers would prefer not to do it
* jansson renames its symbols to something like `jansson_whatever()`, which is an API and ABI break, and presumably the jansson maintainers would prefer not to do it
* both libraries rename their symbols, as above
* *both* libraries use [versioned symbols](https://sourceware.org/binutils/docs/ld/VERSION.html), at least on Linux, and then distributions rebuild dependent binaries against the release that has versioned symbols

Using versioned symbols would mean that:

[...]

PR #639, opened 2020-06-29
build: Add a symbol version to all exported symbols

smcv:
* build: Add symbol versions to all exported symbols
    
    With this version script, newly-linked binaries that depend on the
    json-c shared library will refer to its symbols in a versioned form,
    preventing their references from being resolved to a symbol of the same
    name exported by json-glib or libjansson if those libraries appear in
    dependency search order before json-c, which will usually result in
    a crash. This is necessary because ELF symbol resolution normally uses
    a single flat namespace, not a tree like Windows symbol resolution.
    At least one symbol (json_object_iter_next()) is exported by all three
    JSON libraries.
    
    Linking with -Bsymbolic is not enough to have this effect in all cases,
    because -Bsymbolic only affects symbol lookup within a shared object,
    for example when json_object_set_serializer() calls
    json_object_set_userdata(). It does not affect calls from external
    code into json-c, unless json-c was statically linked into the
    external caller.
    
    This change will also not prevent code that depends on json-glib or
    libjansson from finding json-c's symbols and crashing; to prevent
    that, a corresponding change in json-glib or libjansson would be needed.
    
    Adding a symbol-version is a backwards-compatible change, but once
    added, removing or changing the symbol-version on a symbol would be an
    incompatible change that requires a SONAME bump.
    
    Resolves: https://github.com/json-c/json-c/issues/621
    (when combined with an equivalent change to libjansson).

---

I am not an expert on json-c or CMake, so please review and test carefully.

`--default-symver` replaced by `--version-script=...` as requested. Instead of only introducing a JSONC_0.15 version, I've created a JSONC_0.14 with the symbols that were in 0.14 (I think I've got them right, but please double-check!), and a JSONC_0.15 that follows it, to illustrate what maintainers will need to do for each new release if this approach to versioned symbols is used.


CMakeLists.txt
```diff
@@ -401,6 +401,9 @@ add_library(${PROJECT_NAME}
 set_target_properties(${PROJECT_NAME} PROPERTIES
     VERSION 5.0.0
     SOVERSION 5)
+if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+    target_link_options(${PROJECT_NAME} PRIVATE LINKER:--default-symver)
```
smcv, line 405: This is failing CI because the CMake used on Travis-CI is too old to understand `target_link_options()`. Sorry, I don't know how to backport this (like I said, not a CMake expert).

karelzak:
I don't think --default-symver is proper solution, it's better to use `-version-info <file>` and in the file define JSONC_0.14 { } namespace, and JSONC_0.15 in the next version etc.

smcv:
If the maintainers of json-c are willing to maintain full symbol-versioning (tracking which version introduced each new symbol) as @karelzak suggests, I can redo this merge request to do that. However, that has an ongoing maintenance requirement (each new symbol added to the library needs to be added to the file listing the symbols), so I am not going to propose that unless the maintainers of json-c will take responsibility for doing so in future releases. Please let me know which.

I deliberately proposed this simpler version in the hope that it would maximize the chance of getting some sort of symbol-versioning, that is at least sufficient to disambiguate between the three JSON libraries, adopted by each of them.

CMakeLists.txt
tpgxyz: @smcv  try this

set_target_properties(${PROJECT_NAME} PROPERTIES
    VERSION 5.0.0
    SOVERSION 5
if("${CMAKE_VERSION}" VERSION_LESS "3.13.0" AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Linux") 
    LINK_FLAGS -Wl,--default-symver )
elseif ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
    )
	target_link_options(${PROJECT_NAME} PRIVATE LINKER:--default-symver)
else()
	)
endif()

hawicz: That looks like it'll just mis-compile the library with old cmake versions, please don't do that.

Also, this should not be hard coded to check for Linux, instead actually verify that the link flag is supported, perhaps like the check for `-Bsymbolic-functions` is handled.  That handling also is an example of how to include a link flag w/o the target_link_options (or add_link_options) support in newer cmake version.

Finally, rather than using `--default-symver`, as I mentioned on #621 I'd rather have a explicit version script.

smcv: > instead actually verify that the link flag is supported

The reason I didn't take that approach is that if someone switches from a toolchain that does support it to a toolchain that doesn't, with a capability-based check they'll silently get a json-c whose ABI is incompatible with their previous build (going from versioned symbols to no versioned symbols is an incompatible change).

This is probably only a theoretical concern, because I tried a toolchain from 2012 and it already supported this; but if that downgrade does somehow happen, the question is whether you want a build failure that makes it obvious that something is wrong (as implemented here), or a successful build that silently drops the versioned symbols (which is what you'd get with a capability check).

If you're OK with that failure mode being possible, then yes I can take the same approach as for `-Bsymbolic-functions`.

smcv:
json-c seems to have quite a few symbols that aren't in the `json_` namespace. Were these intended to be exported ABI, or were some of them intended to be private?

If they were intended to be private, adding a version-script is an opportunity to hide them from the ABI. However, I don't use or know this library myself (I'm just trying to solve crashes caused by its symbols colliding with other JSON libraries) so I don't really have the necessary context to choose which ones were meant to be exported.

The conservative option is to export everything that was historically exported, even if it was meant to be private, so that's what I'll do for now. Alternatives would be to exclude the private symbols from the ABI, or put them in their own pseudo-version like `JSONC_PRIVATE`.

smcv:
--default-symver replaced by --version-script=... as requested. Instead of only introducing a JSONC_0.15 version, I've created a JSONC_0.14 with the symbols that were in 0.14 (I think I've got them right, but please double-check!), and a JSONC_0.15 that follows it, to illustrate what maintainers will need to do for each new release if this approach to versioned symbols is used.

CMakeLists.txt
hawicz: I think the opportunity to automatically include versioned symbols when the toolchain supports it, without needlessly limiting it to just Linux OSes, outweighs the chance of someone switching toolchains.

smcv: OK, up to you. That approach is implemented in the latest version of this PR.

hawicz:
You're right about there being more symbols than just the json_* ones.  In an ideal world, we'd hide those (or rename them to json_<something> to make it clearer they're meant to be exported), but doing so will require a major version bump, so for now we need to export them as-is.  Putting them under a JSONC_PRIVATE section sounds like a good compromise.

The symbols that should not be exported at all include:
* _json_c_set_last_err
* _json_c_strerror
* _json_c_strerror_enable
* symbols like __bss_start, _end, _fini, etc...
* i.e. anything starting with an underscore

The ones to list as "private" are:
* json_hex_chars (it's in json_object_private.h now, but it used to be in json_object.h)
* mc_* (internal debugging functions)
* printbuf_* (even though json_object_double_to_json_string() references struct printbuf, json-c library users aren't intended to work with an actual printbuf object)
* mutating array_list_* symbols (add, del_idx, free, new, new2, put_idx, shrink and sort)
    * array_list_{bsearch,get_idx,length} should be exported due to json_object_get_array()
* json_parse_{double,int64,uint64}
* mutating lh_table_* symbols (delete, delete_entry, free, insert, insert_w_hash, new, resize)
    * lh_table_{length,lookup*} should be exported due to json_object_get_object()

All other symbols need to be exported as normal versioned symbols. 

I'll cross-check that against you changes a bit later, when I have a bit more time ($dayjob calls).

karelzak:
Thanks guys that you want to resolve this issue seriously, and you're open for new things! ;-)

(JSON-C is an important piece of Linux distros as LUKS2 uses JSON in on-disk headers now.)

hawicz:
The "lh_table_lookup" symbol shouldn't be in the list at all, and I still want to adjust the list based on my previous comment, but I'm merging as-is for now.

Issue #638, opened 2020-06-29
json-c/json_object.c:50:2: error: #error Unable to determine size of ssize_t


MarcT512:
With latest master, make fails with "json-c/json_object.c:50:2: error: #error Unable to determine size of ssize_t"
The last commit that builds for me is around commit #4a546e7 - and the problem occurred sometime after that.

FYI - I can see that config.h has:
`#define SIZEOF_SSIZE_T 8`

...so I'm not sure why this is happening.

**Steps To Reproduce**
Build master on 64bit Linux 

**Version and Platform**
- json-c version: 34334e5d3f89fe7c1e07e776761f5fa614e67fb7
- OS: Slackware Linux 14.2 64bit & Slackware-current 64bit
- Custom cmake/build flags, if any: None

Full log:
```
marc@deepthought:~/git/json-c/build$ make
Scanning dependencies of target json-c
[  1%] Building C object CMakeFiles/json-c.dir/arraylist.c.o
[  2%] Building C object CMakeFiles/json-c.dir/debug.c.o
[  3%] Building C object CMakeFiles/json-c.dir/json_c_version.c.o
[  5%] Building C object CMakeFiles/json-c.dir/json_object.c.o
/home/marc/git/json-c/json_object.c:50:2: error: #error Unable to determine size of ssize_t
   50 | #error Unable to determine size of ssize_t
      |  ^~~~~
/home/marc/git/json-c/json_object.c: In function â_json_object_new_stringâ:
/home/marc/git/json-c/json_object.c:1269:13: error: âSSIZE_T_MAXâ undeclared (first use in this function); did you mean âSSIZE_MAXâ?
 1269 |  if (len > (SSIZE_T_MAX - (sizeof(*jso) - sizeof(jso->c_string)) - 1))
      |             ^~~~~~~~~~~
      |             SSIZE_MAX
/home/marc/git/json-c/json_object.c:1269:13: note: each undeclared identifier is reported only once for each function it appears in
/home/marc/git/json-c/json_object.c: In function â_json_object_set_string_lenâ:
/home/marc/git/json-c/json_object.c:1332:13: error: âSSIZE_T_MAXâ undeclared (first use in this function); did you mean âSSIZE_MAXâ?
 1332 |  if (len >= SSIZE_T_MAX - 1)
      |             ^~~~~~~~~~~
      |             SSIZE_MAX
[...]

PR #640, opened 2020-06-30
Fix build issues with SSIZE_MAX on 64bit Linux

MarcT512:
Fix build issues with SSIZE_MAX on 64bit Linux
Fixes https://github.com/json-c/json-c/issues/638

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -39,13 +39,13 @@
 #error "The long long type isn't 64-bits"
 #endif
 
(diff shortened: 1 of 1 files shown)


MarcT512:
There was another issue in the existing code in json_object.c. UINT_MAX, ULONG_MAX, ULLONG_MAX shouldn't be used to define the max size of a ssize_t, since these are unsigned maxima and ssize_t is a signed type. This is also fixed here.

PR #651, opened 2020-07-30
Getrandom

alanc:
Use `getrandom()` if available in `json_c_get_random_seed()` - it's  lower overhead than opening & reading from /dev/urandom, and works in chroots and other situtations where /dev/urandom is not available. Falls back to existing methods when kernel doesn't support the syscall.

Also adds a simple test for `json_c_get_random_seed()`.


tests/test_random_seed.c
```diff
@@ -0,0 +1,27 @@
+#include <assert.h>
```
hawicz, line 1: Let's drop this test.  It will not be reliable (it's perfectly valid for two random numbers to be identical), and to make it compile & link properly would require exporting json_c_get_random_seed() as a public API, which we should not do.

alanc: Okay, it served my purpose of making sure the new code worked.

PR #659, opened 2020-08-16
fix json_object_get_boolean() to behave like documented

### LINKED ISSUE #658: json_object_get_boolean() returns wrong result for objects and arrays
**Describtion of the bug**
The current implementation of `json_object_get_boolean(jobj)` returns 0 (false) when `jobj` is a `json_type_object` or `json_type_array`.
This seems to be in contrast to the expected behaviour as documented, which is:

> The type is coerced to a json_bool if the passed object is not a json_bool.
>  integer and double objects will return 0 if there value is zero
>  or 1 otherwise. If the passed object is a string it will return
>  1 if it has a non zero length. **If any other object type is passed
>  1 will be returned if the object is not NULL.**

**Steps To Reproduce**

    #include <json-c/json.h>
    #include <stdio.h>

    int main()
    {
      struct json_object* jobj = json_object_new_object();
      int ob = json_object_get_boolean(jobj);
      printf("bool of a json_type_object = %d (should be 1)\n", ob);

      struct json_object* jarr = json_object_new_array();
      int ab = json_object_get_boolean(jarr);
      printf("bool of a json_type_array = %d (should be 1)\n", ab);
    }

Output of this is:

    bool of a json_type_object = 0 (should be 1)
    bool of a json_type_array = 0 (should be 1)

**Version and Platform**
- json-c version: json-c-0.15 (installed via brew)
- OS: macOS 10.15.5

plan44:
Fixes #658 

`json_object_get_boolean()` must return true for objects and arrays

This boils down to: any object type not checked specifically in the switch statement should return 1, not 0.

I know that `null`values are usually represented as having no object at all, which is checked right after the entry into json_object_get_boolean().
However as I am not 100% sure an object having explicit json_type_null type set is impossible to exist under any circumstances, I added a switch case for that, too.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -702,7 +702,8 @@ json_bool json_object_get_boolean(const struct json_object *jso)
 		}
 	case json_type_double: return (JC_DOUBLE_C(jso)->c_double != 0);
 	case json_type_string: return (JC_STRING_C(jso)->len != 0);
-	default: return 0;
+	case json_type_null: return 0;
+	default: return 1;
 	}
 }
 


hawicz:
This is s doc bug, not a code bug, so I'm not merging this..  See #658.

PR #660, opened 2020-08-22
Validate size arguments in arraylist functions.

stoeckmann:
The array_list_new2 function, which is externally reachable through
json_object_new_array_ext, does not check if specified initial size
actually fits into memory on 32 bit architectures.

It also allows negative values, which could lead to an overflow on these
architectures as well. I have added test cases for these situations.

While at it, also protect array_list_shrink against too large
empty_slots argument. No test added because it takes a huge length
value, therefore a lot of items within the array, to overflow the
calculation. In theory this affects 64 bit sytems as well, but since the
arraylist API is not supposed to be used by external applications
according to its header file, the call is protected due to int
limitation of json_object_array_shrink.

### DIFF
--- a/tests/test1.c
+++ b/tests/test1.c
@@ -1,4 +1,5 @@
 #include <assert.h>
+#include <limits.h>
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


tests/test1.c
```diff
@@ -307,6 +308,25 @@ int main(int argc, char **argv)
 	}
 	printf("my_object.to_string()=%s\n", json_object_to_json_string(my_object));
 
+	json_object_put(my_array);
```
hawicz, line 311: I think you end up with a double free here, or rather down below on line 334.

stoeckmann: Thanks for pointing out, fixed.

PR #665, opened 2020-08-22
Handle more allocation failures in json_tokener* functions

stoeckmann:
The allocation of printbuffer could fail. Also avoid strdup for
setlocale return value to prevent allocation issues in the first place.

### DIFF
--- a/json_tokener.c
+++ b/json_tokener.c
@@ -134,6 +134,12 @@ struct json_tokener *json_tokener_new_ex(int depth)
 		return NULL;
 	}
 	tok->pb = printbuf_new();
+	if (!tok->pb)
+	{
+		free(tok);
+		free(tok->stack);
+		return NULL;
+	}
 	tok->max_depth = depth;
 	json_tokener_reset(tok);
 	return tok;


json_tokener.c
```diff
@@ -298,13 +304,8 @@ struct json_object *json_tokener_parse_ex(struct json_tokener *tok, const char *
...
 	}
 #elif defined(HAVE_SETLOCALE)
-	{
-		char *tmplocale;
-		tmplocale = setlocale(LC_NUMERIC, NULL);
-		if (tmplocale)
-			oldlocale = strdup(tmplocale);
-		setlocale(LC_NUMERIC, "C");
-	}
+	oldlocale = setlocale(LC_NUMERIC, NULL);
```
hawicz, line 307: I'm pretty sure this doesn't actually work w/o making a copy.  The strdup is needed to avoid using memory that is no longer valid after the subsequent setlocale() calls.

stoeckmann: Yes, you are right. Wow, there are quite some programs out there which do it wrong then. Time to fix these programs and maybe recommend a re-phrasing of the setlocale manual page to clarify this. Obviously removed the commit from my pull request.

PR #667, opened 2020-09-11
Fixed test1 regression.

stoeckmann:
SIZEOF_SIZE_T might be only defined in config.h.

Include config.h for these systems to pass tests which are only
supposed to be run on 32 bit systems.

Fixes issue #666.


hawicz:
Thanks!

PR #679, opened 2020-11-05
Let json-c be used with obsolete compilers

GitMensch:
adding stdint typedefs

and if you like to, you could add

```c
#if defined (_MSC_VER) && _MSC_VER < 1600
#define MISSING_STDINT_H
#endif
```

up front...

Feel free to close this PR if somewhere it is explicit said that the compiled json-c library may not be used with that old compilers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,11 +143,14 @@ check_include_file(sys/random.h     HAVE_SYS_RANDOM_H)
 check_include_file(sys/stat.h       HAVE_SYS_STAT_H)
 check_include_file(xlocale.h        HAVE_XLOCALE_H)
 
+# Set json-c specific vars to stamp into json_config.h
+# in a way that hopefully won't conflict with other
+# projects that use json-c.
 if (HAVE_INTTYPES_H)
-	# Set a json-c specific var to stamp into json_config.h
-	# in a way that hopefully won't conflict with other
-	# projects that use json-c.
-    set(JSON_C_HAVE_INTTYPES_H 1)
+	set(JSON_C_HAVE_INTTYPES_H 1)
+endif()
+if (HAVE_STDINT_H)
+	set(JSON_C_HAVE_STDINT_H 1)
 endif()
 
 check_symbol_exists(_isnan          "float.h" HAVE_DECL__ISNAN)
--- a/cmake/json_config.h.in
+++ b/cmake/json_config.h.in
@@ -1,2 +1,5 @@
 /* Define to 1 if you have the <inttypes.h> header file. */
 #cmakedefine JSON_C_HAVE_INTTYPES_H @JSON_C_HAVE_INTTYPES_H@
+
+/* Define to 1 if you have the <stdint.h> header file. */
+#cmakedefine JSON_C_HAVE_STDINT_H @JSON_C_HAVE_STDINT_H@
--- a/json_inttypes.h
+++ b/json_inttypes.h
@@ -13,7 +13,15 @@
 #include <inttypes.h>
 
 #else
+#ifdef JSON_C_HAVE_STDINT_H
 #include <stdint.h>
+#else
+/* Really only valid for old MS compilers, VS2008 and earlier: */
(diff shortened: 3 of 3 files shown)


hawicz:
This appears to be to try to compile on Visual Studio 2008 or earlier.  Where are you trying to build that you need such an old version?

I think the earliest json-c should officially try to support is VS2010, and even that is stretching things since the 2010 version is already [out of support](https://docs.microsoft.com/en-us/visualstudio/productinfo/vs2017-servicingarchive-vs).  However, given the caveat that we will not in any way guarantee that builds will continue to work, I'm willing to merge minimally intrusive, well designed changes that allow things to work for you.  In this case, that means:

 * Use the [already determined](https://github.com/json-c/json-c/blob/6cf48477960b96aedca2c87cf7bb53861ceeecd2/CMakeLists.txt#L143) HAVE_STDINT_H define.
 * Add a [check_type_size](https://cmake.org/cmake/help/v3.0/module/CheckTypeSize.html) (which includes an existence check) for the __int32 and __int64 types before using them.

GitMensch:
The issue is that this is a library-used header - so you cannot use anything that is defined by CMAKE.
If the header would use these defines then every project that currently includes json.h would need to be adjusted. The suggested patch does not break anything and allows the user, if he use an version that does not provide stdint, to manually use the workaround.

As I've said: I'm fine with an explicit note (which I did not found, but I've only did a quick check) that specifies that a newer version is necessary to build and use json-c (I haven't testing building on this, just used a packaged json-c,dl+lib+header, which worked fine after this single header change.

micahsnyder:
I believe json-c's method for defining standard types via `json_inttypes.h` is broken in general. I don't think this PR is the correct solution.

I'll explain: Though `json_inttypes.h` claims to be "internal", it is included when you `#include <json.h>` because `json.h` includes `json_object.h` and `json_object.h` include `json_inttypes.h`. Because `JSON_C_HAVE_INTTYPES_H` isn't defined for downstream projects, those projects will always use the `#else` clause when including `json.h`.  

A hack to make it work properly in downstream projects as-is is to detect inttypes.h and define `JSON_C_HAVE_INTTYPES_H` in downstream projects' config.h files, like so: 
- Detect: https://github.com/micahsnyder/clamav-devel/blob/CLAM-34-unit-test-overhaul-windows-support/CMakeLists.txt#L402
- Define: https://github.com/micahsnyder/clamav-devel/blob/CLAM-34-unit-test-overhaul-windows-support/clamav-config.h.cmake.in#L182

To make it so projects which use json-c don't need to define `JSON_C_HAVE_INTTYPES_H`, json-c could generate `json_inttypes.h` from a `json_inttypes.h.in` file at configure-time much like we did with `clamav-types.h.in` -> `clamav-types.h`: https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.104/clamav-types.h.in

hawicz:
Closing, since the changes as they are don't address the comments above.

GitMensch:
> To make it so downstream projects don't need to define `JSON_C_HAVE_INTTYPES_H` on behalf of json-c, json-c could generate `json_inttypes.h` from a `json_inttypes.h.in` file at configure-time much like we did with `clamav-types.h.in` -> `clamav-types.h`: https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.104/clamav-types.h.in

That sounds like the right solution - should I create an issue for that or is this already in the backlog?

hawicz:
I don't think we need that, it's already done: json_config.h is generated (from json_config.h.in) and installed as a public header, and that's where json_inttypes.h gets the define from.

GitMensch:
Until now it only gets the define to either use inttypes.h or stdint.h; this PR is about a system which does not provide either.

hawicz:
Ok, fine, but even so it should use the json_config.h approach and a `JSON_C_HAVE_STDINT_H` define.

GitMensch:
please reopen and possibly set to draft; I could then have a look how to do adjust json_config (just unlikely to happen in the next weeks by myself)

PR #694, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks !=.
I suggest a simple fix.


hawicz:
Good catch, thanks!  I'm surprised gcc's warnings didn't flag this, but oh well.

PR #701, opened 2021-04-16
[RFC] json_pointer: allow the feature to be disabled

commodo:
Some users may not want to included it in their build/system. So allow a
cmake symbol to disable it.

A user can do 'cmake -DDISABLE_JSON_POINTER=ON <json_c_root_dir>' and
disable the json_pointer functionality. That saves about 17 KB (on an
x86_64) machine. This may be useful on smaller embedded systems; even
though the saving would be fewer kilobytes.

One thing that also needs to change a bit, is that the 'json.h' be
autogenerated via cmake, in order to conditionally include that
"json_pointer.h" file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,7 @@
 # It's not good practice to build directly in the source tree
 # but ignore cmake auto-generated files anyway:
 /json_config.h
+/json.h
 /config.h
 /json-c.pc
 /Makefile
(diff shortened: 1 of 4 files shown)


commodo:
As a note here: I admit that doing this disable knob may open a can of worms in terms of making plenty of of other disable knobs. Or adding a new feature would maybe require a disable knob (if it's not part of the standard JSON core functionality).

But I thought I'd try this PR as an RFC.
[btw: I implemented JSON pointer]
I was looking at some old code, and found some basic C for JSON patch, and wanted to take a stab at finishing it.

The fact that it's all cmake now is nice. I don't need to modify a CMakeLists.txt + Makefile.am + VisualStudio project file anymore (so yay :D )

commodo:
Changelog v1 -> v2:
* removed commented line in CMakeLists.txt

commodo:
Changelog v2 -> v3:
* indented blocks of code under if() statements with 4 spaces;

michellemartni:
Stop sending me this thanks

commodo:
Well, you ca unsubscribe from this thread.

There’s a button for that on this page.

Or you can choose to unwatch this repo.

I don’t have control over these options for your Github account.

PR #702, opened 2021-04-26
json_patch: add first implementation only with patch application

commodo:
Initially I wanted to also do a function that generates the JSON patch from
two JSON documents, but even just applying the JSON patch was a bit of
work, especially when needing to satisfy all the test-cases.

This change defines all the operation in the RFC6902. The addition isn't
too big (for the json_patch_apply() function), as part of the heavy lifting
is also done by JSON pointer logic.

All the ops were tested with the test-cases defined at:
  https://github.com/json-patch/json-patch-tests

RFC6902: https://tools.ietf.org/html/rfc6902

However, in order to properly support JSON patch:
* json_object_array_insert_idx() is needed in the JSON object API
* this changeset also changes [slightly] the behavior of `json_object_set{f}()` ; I figured, it may be still doable; I can revert that behavior change [if needed]; the behavior of `json_object_set{f}()` wasn't clarified initially;
* the `json_pointer_set_with_array_cb()` mechanism should be useful to support various JSON array behavior types

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -403,6 +403,7 @@ set(JSON_C_PUBLIC_HEADERS
 set(JSON_C_HEADERS
     ${JSON_C_PUBLIC_HEADERS}
     ${PROJECT_SOURCE_DIR}/json_object_private.h
+    ${PROJECT_SOURCE_DIR}/json_pointer_private.h
     ${PROJECT_SOURCE_DIR}/random_seed.h
     ${PROJECT_SOURCE_DIR}/strerror_override.h
     ${PROJECT_SOURCE_DIR}/math_compat.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,8 @@ option(ENABLE_RDRAND                  "Enable RDRAND Hardware RNG Hash Seed."   
 option(ENABLE_THREADING               "Enable partial threading support."                     OFF)
 option(OVERRIDE_GET_RANDOM_SEED       "Override json_c_get_random_seed() with custom code."   OFF)
 option(DISABLE_EXTRA_LIBS             "Avoid linking against extra libraries, such as libbsd." OFF)
-option(DISABLE_JSON_POINTER           "Disable JSON pointer (RFC6901) support."               OFF)
+option(DISABLE_JSON_POINTER           "Disable JSON pointer (RFC6901) and JSON patch support." OFF)
+option(DISABLE_JSON_PATCH             "Disable JSON patch (RFC6902) support."                 OFF)
 option(NEWLOCALE_NEEDS_FREELOCALE     "Work around newlocale bugs in old FreeBSD by calling freelocale"  OFF)
 
 
@@ -429,8 +430,15 @@ if (NOT DISABLE_JSON_POINTER)
     set(JSON_C_PUBLIC_HEADERS   ${JSON_C_PUBLIC_HEADERS}  ${PROJECT_SOURCE_DIR}/json_pointer.h)
     set(JSON_C_SOURCES          ${JSON_C_SOURCES}         ${PROJECT_SOURCE_DIR}/json_pointer.c)
     set(JSON_H_JSON_POINTER "#include \"json_pointer.h\"")
+
+    if (NOT DISABLE_JSON_PATCH)
+        set(JSON_C_PUBLIC_HEADERS   ${JSON_C_PUBLIC_HEADERS}  ${PROJECT_SOURCE_DIR}/json_patch.h)
+        set(JSON_C_SOURCES          ${JSON_C_SOURCES}         ${PROJECT_SOURCE_DIR}/json_patch.c)
+	set(JSON_H_JSON_PATCH "#include \"json_patch.h\"")
+    endif()
 else()
     set(JSON_H_JSON_POINTER "")
+    set(JSON_H_JSON_PATCH "")
 endif()
 
 configure_file(json.h.cmakein ${PROJECT_BINARY_DIR}/json.h @ONLY)
--- a/arraylist.c
+++ b/arraylist.c
@@ -125,6 +125,27 @@ int array_list_shrink(struct array_list *arr, size_t empty_slots)
 	return 0;
 }
 
+int array_list_insert_idx(struct array_list *arr, size_t idx, void *data)
+{
+	size_t move_amount;
+
+	if (idx >= arr->length)
+		return array_list_put_idx(arr, idx, data);
+
+	/* we're at full size, what size_t can support */
+	if (arr->length == SIZE_T_MAX)
+		return -1;
+
+	if (array_list_expand_internal(arr, arr->length + 1))
+		return -1;
+
+	move_amount = (arr->length - idx) * sizeof(void *);
+	memmove(arr->array + idx + 1, arr->array + idx, move_amount);
+	arr->array[idx] = data;
+	arr->length++;
+	return 0;
+}
+
 //static inline int _array_list_put_idx(struct array_list *arr, size_t idx, void *data)
 int array_list_put_idx(struct array_list *arr, size_t idx, void *data)
 {
--- a/json-c.sym
+++ b/json-c.sym
@@ -18,6 +18,7 @@ JSONC_PRIVATE {
     array_list_new;
     array_list_put_idx;
     array_list_sort;
+    json_pointer_get_internal;
     json_hex_chars;
     json_parse_double;
     json_parse_int64;
--- a/json-c.sym
+++ b/json-c.sym
@@ -176,4 +176,5 @@ JSONC_0.16 {
 JSONC_0.17 {
 #  global:
 #      ...new symbols here...
+#    array_list_insert_idx is intentionally not exported
 } JSONC_0.16;
--- a/json-c.sym
+++ b/json-c.sym
@@ -167,8 +167,8 @@ JSONC_0.15 {
 } JSONC_0.14;
 
 JSONC_0.16 {
-#  global:
-#      ...new symbols here...
+  global:
+    json_object_array_insert_idx;
 } JSONC_0.15;
 
 JSONC_0.17 {
--- a/json-c.sym
+++ b/json-c.sym
@@ -170,6 +170,7 @@ JSONC_0.15 {
 JSONC_0.16 {
   global:
     json_object_array_insert_idx;
+    json_patch_apply;
 } JSONC_0.15;
 
 JSONC_0.17 {
--- a/json-c.sym
+++ b/json-c.sym
@@ -18,7 +18,6 @@ JSONC_PRIVATE {
     array_list_new;
     array_list_put_idx;
     array_list_sort;
-    json_pointer_get_internal;
     json_hex_chars;
     json_parse_double;
     json_parse_int64;
@@ -168,13 +167,12 @@ JSONC_0.15 {
 } JSONC_0.14;
 
 JSONC_0.16 {
-  global:
-    json_object_array_insert_idx;
-    json_patch_apply;
+# No new symbols in 0.16
 } JSONC_0.15;
 
 JSONC_0.17 {
-#  global:
-#      ...new symbols here...
+  global:
+    json_object_array_insert_idx;
+    json_patch_apply;
 #    array_list_insert_idx is intentionally not exported
 } JSONC_0.16;
--- a/json_object.h
+++ b/json_object.h
@@ -622,6 +622,25 @@ JSON_EXPORT int json_object_array_add(struct json_object *obj, struct json_objec
 JSON_EXPORT int json_object_array_put_idx(struct json_object *obj, size_t idx,
                                           struct json_object *val);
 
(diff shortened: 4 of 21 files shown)


commodo:
I can split this up in smaller PRs to reduce the amount of review effort.
Some changes in this set can be moved independently, if that helps.

tests/test_json_patch.c
```diff
@@ -0,0 +1,119 @@
...
+		res = NULL;
+	}
+
+	printf("OK\n");
+}
+
+void test_json_patch_using_file(const char *testdir, const char *filename)
+{
+	char full_filename[PATH_MAX];
+	(void)snprintf(full_filename, sizeof(full_filename), "%s/%s", testdir, filename);
```
hawicz, line 72: You need a #include "snprintf_compat.h" at the top of this file to use snprintf.  (otherwise Windows builds break)

arraylist.c
```diff
@@ -125,6 +125,27 @@ int array_list_shrink(struct array_list *arr, size_t empty_slots)
 	return 0;
 }
 
+int array_list_insert_idx(struct array_list *arr, size_t idx, void *data)
+{
+	size_t move_amount;
+
+	if (idx > arr->length)
```
hawicz, line 132: This should be "if (idx >= arr->length)"

json-c.sym
```diff
@@ -16,8 +16,10 @@ JSONC_PRIVATE {
     array_list_del_idx;
     array_list_free;
     array_list_new;
+    array_list_insert_idx;
```
hawicz, line 19: Please do not add anything to JSONC_PRIVATE.  Unlike the previously existing symbols, if these new ones are internal-only, they shouldn't be exported at all.

json_object.h
```diff
@@ -605,6 +605,27 @@ JSON_EXPORT int json_object_array_add(struct json_object *obj, struct json_objec
 JSON_EXPORT int json_object_array_put_idx(struct json_object *obj, size_t idx,
                                           struct json_object *val);
 
+/** Insert an element at a specified index in an array (a json_object of type json_type_array)
+ *
+ * The reference count will *not* be incremented. This is to make adding
+ * fields to objects in code more compact. If you want to retain a reference
+ * to an added object you must wrap the passed object with json_object_get
+ *
+ * The reference count of a replaced object will be decremented.
```
hawicz, line 614: This func won't replace any object; delete this line.

json_object_private.h
```diff
@@ -100,6 +100,25 @@ void _json_c_set_last_err(const char *err_fmt, ...);
 
 extern const char *json_hex_chars;
 
+struct json_pointer_get_result {
```
hawicz, line 103: Let's add a json_pointer_private.h, to keep these a little better contained to just json_patch.c and json_pointer.c

json_patch.c
```diff
@@ -0,0 +1,249 @@
...
+{
+	json_pointer_array_set_cb array_set_cb;
+	struct json_pointer_get_result from;
+	struct json_object *jfrom;
+	const char *from_s;
+	size_t from_s_len;
+	int rc;
+
+	if (!json_object_object_get_ex(patch_elem, "from", &jfrom)) {
+		errno = EINVAL;
```
hawicz, line 148: Let's avoid (ab)using errno to return additional json-c-specific error information.  Instead, consider calling _json_c_set_last_err() with a more detailed message.

json_patch.c
```diff
@@ -0,0 +1,249 @@
...
+	size_t i;
+	int rc = 0;
+
+	if (!base || !json_object_is_type(patch, json_type_array)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	/* errno should be set inside json_object_deep_copy() */
+	if (json_object_deep_copy(base, res, NULL) < 0)
```
hawicz, line 205: It should be up to the caller to copy the incoming object or not.  I can imagine that in many cases creating a copy will be needless overhead.

json_patch.h
```diff
@@ -0,0 +1,46 @@
...
+extern "C" {
+#endif
+
+/**
+ * Apply the JSON patch to the base object.
+ * The patch object must be formatted as per RFC 6902.
+ * If the patch is not correctly formatted, an error will
+ * be returned.
+ *
+ * The original `base` object will first be copied, and then
```
hawicz, line 28: As I mentioned on the json_object_deep_copy() call in json_patch.c, let's leave this up to the caller to do or not.

json_patch.c
```diff
@@ -0,0 +1,249 @@
...
+	if (rc)
+		json_object_put(from.obj);
+
+	return rc;
+}
+
+int json_patch_apply(struct json_object *base, struct json_object *patch,
+                     struct json_object **res)
+{
+	size_t i;
```
hawicz, line 196: please don't use single character variable named.  e.g. call this "ii" instead.

json_patch.c
```diff
@@ -0,0 +1,249 @@
...
+
+static int __json_patch_apply_remove(struct json_pointer_get_result *jpres)
+{
+	if (json_object_is_type(jpres->parent, json_type_array)) {
+		return json_object_array_del_idx(jpres->parent, jpres->id.index, 1);
+	} else if (jpres->parent && jpres->id.key) {
+		json_object_object_del(jpres->parent, jpres->id.key);
+		return 0;
+	} else {
+		return json_object_put(jpres->obj);
```
hawicz, line 56: Is this the case where you're removing the root object?  If so, doesn't that mean that *res is now invalid, and should be set to NULL?  Is removing the root even a valid thing to do?

json_patch.c
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	rc = json_pointer_get_internal(*res, from_s, &from);
+	if (rc)
+		return rc;
+
+	json_object_get(from.obj);
+
+	if (!move) {
+		array_set_cb = json_object_array_insert_idx_cb;
```
hawicz, line 176: What if from.parent isn't an array?  If you're not supporting a parent that's a json_type_object then you should check the type here, before getting as far as calling any json_object_array* functions on the wrong object type.  (and note the limitation in the json_patch_apply() docs)

json_patch.c
hawicz: Also, it would be worthwhile adding a json patch specific error code param to the json_patch_apply() signature, to ensure it's possible for callers to implement the recommended [error handling](https://tools.ietf.org/html/rfc5789#section-2.2) based on how things actually failed.

commodo:
it will take me a while to re-spin everything; i came back from a short vacation;
would it be ok to do a PR with just the first 3 patches?
i'll probably do a PR with just those 3 in the meantime;

arraylist.c
commodo: the code below seems to work fine to cover the `idx == arr->length`;
but it makes sense to do `idx >= arr->length` here

so, ack

json-c.sym
commodo: ack

json_object.h
commodo: ack;
this was silly of me

commodo:
Changelog v1 -> v2:
* only first 3 commits changed [for now] ; also moved to PR https://github.com/json-c/json-c/pull/704 in case that simplifies things; i'm also fine to close that
* https://github.com/json-c/json-c/compare/a0c7225cd907a1cf634a30584c3bb2a1531170fa..5ff17b986defe4fc0f604eb141a876fdf67e46be
* changed `idx > arr->length` -> `idx >= arr->length`
* removed `array_list_insert_idx` from `JSONC_PRIVATE`
* removed ` The reference count of a replaced object will be decremented.` docstring from `json_object_array_insert_idx()` ; the patch may show up like it's `json_object_array_put_idx()`
* updated test-case for json_object_array_insert_idx() to insert last element as well

hawicz:
Fyi, I adjusted some of these changes, and I'm adding additional fixes over at https://github.com/hawicz/json-c/tree/json_patch

tests/test_json_patch.c
hawicz: done in https://github.com/hawicz/json-c/tree/json_patch

json_object_private.h
hawicz: done in https://github.com/hawicz/json-c/tree/json_patch

json_patch.c
hawicz: fixed in my fork

json_patch.c
hawicz: I'm working on this in my fork

json_patch.c
hawicz: I'm working on this in my fork

commodo:
> Fyi, I adjusted some of these changes, and I'm adding additional fixes over at https://github.com/hawicz/json-c/tree/json_patch

Awesome :)
Thanks

Apologies I did not get a chance to do any updates on this.
Hopefully, all the work (I did) helps.
I'm having a bit of a tough time remembering implementation details on this; but I'll try to take a look.

hawicz:
Hang tight for a bit, I've got a number of local changes that I'll be pushing to my fork which address my earlier review comments.  (it might take a day or two)

json_patch.c
hawicz: I added a `struct json_patch_error` arg to `json_patch_apply()`.

json_patch.c
hawicz: I've changed the meaning of the args a bit, there is now a `copy_from` arg and a `base` arg.

json_patch.c
hawicz: Fixed at https://github.com/hawicz/json-c/commit/670bf12f9ee5cef46a116c3350b78cf4a358ec2b

json_patch.c
hawicz: Never mind, I didn't quite understand how this works.  There is a check inside json_pointer for the type of the object, and only when it _is_ an array is this callback used, so this is fine.

hawicz:
@commodo, I believe this is in a pretty good shape to merge now.  Would you care to give it a once over before I do so?

CMakeLists.txt
```diff
@@ -88,7 +88,8 @@ option(ENABLE_RDRAND                  "Enable RDRAND Hardware RNG Hash Seed."
 option(ENABLE_THREADING               "Enable partial threading support."                     OFF)
 option(OVERRIDE_GET_RANDOM_SEED       "Override json_c_get_random_seed() with custom code."   OFF)
 option(DISABLE_EXTRA_LIBS             "Avoid linking against extra libraries, such as libbsd." OFF)
-option(DISABLE_JSON_POINTER           "Disable JSON pointer (RFC6901) support."               OFF)
+option(DISABLE_JSON_POINTER           "Disable JSON pointer (RFC6901) and JSON patch support." OFF)
+option(DISABLE_JSON_PATCH             "Disable JSON patch (RFC6902) support."                 OFF)
```
commodo, line 92: nice: thanks for these :)
i've been also meaning to do this;
i was a bit undecided whether to do them in CMake or classic Make

json_patch.c
```diff
@@ -0,0 +1,332 @@
...
+	if (json_pointer_get_internal(*res, path, &jpres))
+	{
+		_set_err_from_ptrget(errno, "path");
+		return -1;
+	}
+
+	rc = __json_patch_apply_remove(&jpres);
+	if (rc < 0)
+		_set_err(EINVAL, "Unable to remove path referenced by 'path' field");
+	// This means we removed and freed the root object, i.e. *res
```
commodo, line 105: did i add the `freed` typo?

commodo (commented):
All looks good to me :)

I'm mostly counting on the json_patch tests to pass.
IIRC: they're taken from some official JSON-patch repo.

json_patch.c
hawicz: No, I wrote that (intentionally past tense)

CMakeLists.txt
hawicz: er... actually _you_ did that: 393c98fb  :) 
I was rather surprised github allowed me to push to a branch in your fork.  Did you explicitly allow that?  
Anyway, it allowed for both of our changes to be rolled into this same PR, at the expense of a bit of confusion over the origin of the code, at least if you're just looking at the overall diffs.

commodo: > Did you explicitly allow that?

no, i went with defaults;

commodo:
Wooo
Thanks :)

I know this sounds a bit ¯\\\_(ツ)\_/¯  , but with this, I'm now finished with json-c (for a while).

Contextually: 
* json-c is used by OpenWrt (as a core lib)
* I was working for a company a few years ago, and we were using OpenWrt in a project
* We were experimenting with solutions to address issues of large JSON files (50-100 MB JSON files).
  * Someone might ask whether JSON is the best format for large data-sets
  * For us, that was the situation between configuration coming from the cloud and having many network devices getting their configs (with routing info and stuff)
* JSON patch was a solution for this, but we needed something fast
  *  I don't know what they ended up doing (I left at some point)
* I only managed to implement JSON pointer (while still working on that project)
* I never thought I'd finish JSON patch; I started it sometime after I left (since it seemed fun to do)

Thanks again :)
Alex

PR #703, opened 2021-05-03
Fix vasprintf fallback

raffenet:
Once a va_list has been processed, it cannot be used again. Make a
copy and use it for the follow-up call to vsprintf.


hawicz:
The AppVeyor build here failed because `va_copy` isn't available on Visual Studio 2010 (the "b_toolset=Windows7.1SDK" build).  Given how old VS2010 is, I'm expecting to [drop support ](https://groups.google.com/g/json-c/c/SdHYnjqXtf4 )for that, at which point we can merge this.

The Travis build failure was just a timeout w/ the osx build, so that can just be ignored.

raffenet:
> The AppVeyor build here failed because `va_copy` isn't available on Visual Studio 2010 (the "b_toolset=Windows7.1SDK" build). Given how old VS2010 is, I'm expecting to [drop support ](https://groups.google.com/g/json-c/c/SdHYnjqXtf4)for that, at which point we can merge this.

I can stick the `va_copy` inside the else branch of `#ifdef WIN32` if that helps. Windows isn't supported in my use-case anyways.

PR #714, opened 2021-07-05
Add AfterCaseLabel to .clang-format

Hex052:
This is to fix the behavior that might've changed between older versions of clang-format, I'm not sure.
Regardless, version 10 (and possibly earlier, I didn't check) tries to put the bracket on the same line as case without this.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -23,6 +23,8 @@ AllowShortFunctionsOnASingleLine:  Empty
 BreakBeforeBraces: Custom
 # Control of individual brace wrapping cases.
 BraceWrapping:
+  # Wrap brackets inside of a case
+  AfterCaseLabel: true
   # Wrap class definition.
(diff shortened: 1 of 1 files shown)


hawicz:
Older versions of clang-format complain with: 
```
YAML:26:19: error: unknown key 'AfterCaseLabel'
  AfterCaseLabel: true
```

I don't suppose there's a way to make this conditional on the clang-format version, is there?

Hex052:
@hawicz I installed clang-format 6 through 11 and it looks like versions 9 and later need `AfterCaseLabel: true` while 8 and earlier do not. This appears to be the only change that affects json-c between those versions, and unfortunately there does not appear to be any way to specify the version based on the version of clang-format.

hawicz:
bleh, oh well, let's move ahead with the newer support

PR #715, opened 2021-07-07
Allow null character in keys

Hex052:
This allows keys to contain null characters (`\u0000`). While there were no changes made to the `json_object_object_*` functions, there were some changes needed for some of the others.

Before merging this, I'd want to go through and rename a fair amount of symbols that are badly named (such as moving and renaming one of the types I added, `struct lh_string`). Before I go through that effort, though, I'd want to know what, if anything, I should change in what I have written.

This tries to fix issue #108 (merge request #528 will no longer be needed)

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -39,18 +39,6 @@
 #error "The long long type isn't 64-bits"
 #endif
 
-#ifndef SSIZE_T_MAX
-#if SIZEOF_SSIZE_T == SIZEOF_INT
-#define SSIZE_T_MAX INT_MAX
-#elif SIZEOF_SSIZE_T == SIZEOF_LONG
-#define SSIZE_T_MAX LONG_MAX
-#elif SIZEOF_SSIZE_T == SIZEOF_LONG_LONG
-#define SSIZE_T_MAX LLONG_MAX
-#else
-#error Unable to determine size of ssize_t
-#endif
-#endif
-
 // Don't define this.  It's not thread-safe.
 /* #define REFCOUNT_DEBUG 1 */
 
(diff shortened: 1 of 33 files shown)


Hex052:
Hmm, well, good to know it doesn't compile on Windows (I should've tested that)

Hex052:
@hawicz This now compiles properly and is ready for review

hawicz:
First of all, thank you for working on this.  I don't have a whole lot of time right now, but the things that I noticed so far from glancing at the diffs are:
* This changes API of json_c_shallow_copy_default and json_object_object_foreach, which means it would require a major version bump.  Can we avoid that?
* New functions that don't follow the naming pattern of preexisting ones, like json_key_size(), etc... would be better if they were named more json-c specific.  e.g. maybe json_c_key_size()?
* It seems like a few of the new API functions could be consolidated, and just provide _key versions that take a json_key.  (and perhaps provide a "json_c_key_make_temp(const char *, ssize_t)")
* However, I'm a bit wary of creating a json-c specific length-based buffer, i.e. struct json_key.  If there's no good pre-existing, widely used similar structure to use, then ok, but I suspect we might be able to make a better choice here.

I'm planning on taking a more in-depth look at this soon, so I'll probably have a few more things to comment on.

json_object.c
```diff
@@ -1775,6 +1875,89 @@ static int json_object_deep_copy_recursive(struct json_object *src, struct json_
 	return 0;
 }
 
+/*
+ * The actual guts of json_object_deep_copy(), with a few additional args
+ * needed so we can keep track of where we are within the object tree.
+ *
+ * Note: caller is responsible for freeing *dst if this fails and returns -1.
+ */
+static int json_object_deep_copy_recursive_len(struct json_object *src, struct json_object *parent,
```
hawicz, line 1884: json_object_deep_copy_recursive_len() looks like a lot of code duplication just to pass a different shallow_copy function pointer.  Instead, add another argument to the existing json_object_deep_copy_recursive() and have it use whichever one is non-NULL.

json_object.h
```diff
@@ -458,29 +597,104 @@ JSON_EXPORT json_bool json_object_object_get_ex(const struct json_object *obj, c
...
+ *
+ * @param obj the json_object instance
+ * @param key the local name for the `char *` key variable defined in the body
+ * @param val the local name for the `json_object *` object variable defined in
+ *            the body
+ */
+#if defined(__GNUC__) && !defined(__STRICT_ANSI__) && \
+    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L)
+
+#define json_object_object_foreach(obj, key, val)                                            \
```
hawicz, line 642: The json_object_object_foreach_len() macro duplicates most of the code for json_object_object_foreach().  How about instead of doing that we have json_object_object_foreach() define a variable in addition to "key", perhaps "key_with_len" (though I'm not too excited about that var name)?

hawicz:
I started gathering some notes about possible buffer implementations to use over at https://github.com/json-c/json-c/wiki/Proposal:-Use-sized-buffers-instead-of-asciiz-strings
Given what I've found so far, there doesn't seem to be anything exactly like what I'm looking for.  However, if we're going with json-c specific code, it just occurred to me that the existing arraylist code already implements a fair amount of the buffer handling, so maybe we could just abuse/extend that slightly.

Also, the naming of the new functions, <foo>_key isn't so great either, because "key" doesn't seem to describe very well what makes those different.

Hex052:
I'm not a huge fan of _key either, but I didn't really know what to name them with. Also, for the `struct json_key`, would it be a good idea to modify `struct json_string` so they can both share some common items? Since that's basically what keys are is just strings.

Hex052:
Apologies for the delay, I just had a concern, am I free to modify `linkhash.h`? I currently have in order to actually be able to meaningfully be able to hash and compare keys with null characters. If the major version is being increased, can we drop `linkhash.h` as a publicly includeable file? Seems like that'd make it easier to make changes in the future, too.

I think it would be *possible* to get away without modifying either that or `json_c_visit_userfunc` (in `json_visit.h`), but it would certainly be really kludgy, involving ensuring the key always contained the data inline `const char *` and prepending data before each of the pointers passed to those functions? I'm not sure how far you'd be willing to go to maintain backward compatibility, though.

Though, I'd already be modifying `json_object_deep_copy_recursive` with your suggestion, so I think it'd need to bump the major version number. I think all the changes to avoid a major version bump would be difficult to maintain.

PR #716, opened 2021-07-09
Fixed cmake command

Frityet:
The directory `json-c` in the root of this repository does not exist, and just running `cmake` in the root works fine


Frityet:
nvm i read the entire command wrong

PR #717, opened 2021-07-23
Cmake is able delete all files by "clean" target

Pawday:
That target name conflicts by same target name in another library (curl)


hawicz:
Sorry, I'm not going to merge this since it just drops support for a "make uninstall" target entirely.  "cmake clean" does _not_ do the same thing.

hawicz:
I added a guard around the uninstall target in commit 8c727e5, hopefully that'll help.

PR #718, opened 2021-07-26
CMake create uninstall target if unix generator is used

Pawday:


hawicz:
Well, ok, we can limit it, but why bother?  Does it cause problems?

Pawday:
That target is not working on Windows with Nmake and VS generators

hawicz:
Well, of course not, I wouldn't expect windows to have "cat", "xargs" and "rm", but if you don't actually run the target, does it matter?

PR #728, opened 2021-10-12
DRAFT PROPOSAL - Add option JSON_C_OBJECT_ADD_IF_NOT_NULL

jobol:
This option of json_object_object_add_ex will cancel the
addition if the added value is NULL.

Remarks:
 - testing value with json_object_is_type(val, json_type_null)
   could make sense
 - that commit introduces a new test that could slow the speed
   of add

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -591,6 +591,9 @@ int json_object_object_add_ex(struct json_object *jso, const char *const key,
 
 	assert(json_object_get_type(jso) == json_type_object);
 
+	if (val == NULL && (opts & JSON_C_OBJECT_ADD_IF_NOT_NULL))
+		return 0;
+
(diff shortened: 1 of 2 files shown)


hawicz:
I'm not convinced that adding this flag to json-c is worth it.  Adding it means yet another flag, tests to verify it works, plus it introduces an inconsistency as compared to json_object_array_add().
It seems like it would be simpler to have callers change their code:


```
if (json_object_object_add_ex(jso, key, val, opts) != 0) {
   ...error...
}
```

to instead do:

```
if (val != NULL && json_object_object_add_ex(jso, key, val, opts) != 0) {
   ...error...
}
```

jobol:
okay, just a proposal. regards

PR #740, opened 2022-01-12
Fix warnings of clang -Wshorten-64-to-32

rouault:


hawicz:
I think at least some (all?) of these warnings are actual problems, and rather than papering over them by casting to int, we need to take adjust the types used.  e.g. `printbuf_memappend()` should take a size_t, `int ret` should be `size_t ret` instead, etc...

That might break API/ABI compatibility, so I'll need to take a closer look at each change.

hawicz:
I fixed some of these issues in commit bdd5e03.  Changes to json_tokener.c and json_object.c are still needed.

PR #744, opened 2022-02-17
Close file on error path.

Kizuna-Meraki:
The file was only be closed when there was no error and
was being left open when there was an error. By moving
the close(fd) statement out of the if-clause, the file
can be close regardless if there is an error or not.
After the file is closed, it can be checked for errors.

shout out to:
[c3h2_ctf](https://twitter.com/c3h2_ctf)


hawicz:
Thanks for the patch!

PR #751, opened 2022-03-03
printbuf: do not allow invalid arguments

stoeckmann:
If invalid arguments are passed to printbuf functions return -1 to
protect printbuf internals.

Shoutout to [C3H2-CTF](https://twitter.com/c3h2_ctf).


stoeckmann:
It is up to you if handling invalid arguments is something json-c should do. This is clearly a user error after all.

PR #758, opened 2022-03-20
Preserve context if out of memory

stoeckmann:
If memory allocation fails in json_c_set_serialization_double_format or
json_object_copy_serializer_data then return with an error value and
preserve previous values without overriding them with NULL.

Shoutout to [C3H2-CTF](https://twitter.com/c3h2_ctf).

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -933,7 +933,21 @@ int json_c_set_serialization_double_format(const char *double_format, int global
 #endif
 		if (global_serialization_float_format)
 			free(global_serialization_float_format);
-		global_serialization_float_format = double_format ? strdup(double_format) : NULL;
+		if (double_format)
+		{
+			char *p = strdup(double_format);
+			if (p == NULL)
+			{
+				_json_c_set_last_err("json_c_set_serialization_double_format: "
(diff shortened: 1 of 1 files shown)


json_object.c
```diff
@@ -933,7 +933,21 @@ int json_c_set_serialization_double_format(const char *double_format, int global
 #endif
 		if (global_serialization_float_format)
 			free(global_serialization_float_format);
-		global_serialization_float_format = double_format ? strdup(double_format) : NULL;
+		if (double_format)
+		{
+			char *p = strdup(double_format);
+			if (p == NULL)
+			{
+				_json_c_set_last_err("json_c_set_set_serialization_double_format: "
```
ploxiln, line 941: typo: doubled `_set_set_`

(also below ... and also back in https://github.com/json-c/json-c/pull/755/files#diff-3dbecd622f44dd0c1d2cc6363ed0f5f5b13d0cc912f1d7580ebb2875dffab023R948 fwiw)

stoeckmann: Fixed, thank you! I have a talent to include typos in a message which I am about to copy...

PR #771, opened 2022-06-02
fix fallthough warning

### LINKED ISSUE #770: json_parse.c:170:13: error: this statement may fall through
/home/ubuntu/json-c/apps/json_parse.c: In function ‘main’:
/home/ubuntu/json-c/apps/json_parse.c:170:13: error: this statement may fall through [-Werror=implicit-fallthrough=]
  170 |   case 'h': usage(argv[0], 0, NULL);
      |             ^~~~~~~~~~~~~~~~~~~~~~~
/home/ubuntu/json-c/apps/json_parse.c:171:3: note: here
  171 |   default: /* '?' */ usage(argv[0], EXIT_FAILURE, "Unknown arguments");
      |   ^~~~~~~
make[2]: *** [apps/CMakeFiles/json_parse.dir/build.make:76: apps/CMakeFiles/json_parse.dir/json_parse.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1698: apps/CMakeFiles/json_parse.dir/all] Error 2
make: *** [Makefile:166: all] Error 2

In case 'h', break is needed to solve this

cosmo-ray:
Hello,

I've try to fix a fall-though error with gcc (11 on my machine).
Instead of having a fall-though comment, as gcc would advice,
I've made an early exit, as using -h is generally expected to just print
help and exit.

note that if you want me to keep old help behavior I can modify this patch too.

fix #770

Signed-off-by: Matthias Gatto <matthias.gatto@protonmail.com>

### DIFF
--- a/apps/json_parse.c
+++ b/apps/json_parse.c
@@ -167,7 +167,7 @@ int main(int argc, char **argv)
 		case 'f': formatted_output = 1; break;
 		case 'n': show_output = 0; break;
 		case 's': strict_mode = 1; break;
-		case 'h': usage(argv[0], 0, NULL);
+		case 'h': usage(argv[0], 0, NULL); exit(EXIT_SUCCESS);
 		default: /* '?' */ usage(argv[0], EXIT_FAILURE, "Unknown arguments");
 		}
 	}


hawicz:
dup of #769, although neither that nor this change are really the right fix, since usage() itself exits.

PR #772, opened 2022-06-02
add JSON_C_TO_STRING_COLOR option

cosmo-ray:
This option enable color in json_object_to_json_string_ext.
I've try to made something similar to jq output,
but I color true/false and null in purple,
as it's what is common color scheme used in programing language in emacs.

also add a '-c' option into json_parser to test it.

note: that I could have done a color() function similar to
what is done with indent(), but as the code is pretty small
I've keep it as it. so if you want  me to use a subfunction
tell me and I'll do it.

Signed-off-by: Matthias Gatto <matthias.gatto@protonmail.com>

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -64,6 +64,12 @@ static void json_object_generic_delete(struct json_object *jso);
 #define inline
 #endif
 
+/* define colors */
+#define ANSI_COLOR_RESET "\033[0m"
+#define ANSI_COLOR_FG_GREEN "\033[0;32m"
+#define ANSI_COLOR_FG_BLUE "\033[0;34m"
+#define ANSI_COLOR_FG_MAGENTA "\033[0;35m"
+
 /*
  * Helper functions to more safely cast to a particular type of json_object
  */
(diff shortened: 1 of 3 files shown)


json_object.c
```diff
@@ -629,9 +642,18 @@ void json_object_object_del(struct json_object *jso, const char *key)
...
+		printbuf_strappend(pb,"\033[0;35m");
+
 	if (JC_BOOL(jso)->c_boolean)
-		return printbuf_strappend(pb, "true");
-	return printbuf_strappend(pb, "false");
+		ret = printbuf_strappend(pb, "true");
+	else
+		ret = printbuf_strappend(pb, "false");
+	if (flags & JSON_C_TO_STRING_COLOR)
+		return printbuf_strappend(pb,"\033[0m");
```
hawicz, line 655: If the previous printbuf_strappend() returned failure, we should not call this one.

json_object.c
```diff
@@ -474,15 +474,28 @@ static int json_object_object_to_json_string(struct json_object *jso, struct pri
 		if (flags & JSON_C_TO_STRING_SPACED && !(flags & JSON_C_TO_STRING_PRETTY))
 			printbuf_strappend(pb, " ");
 		indent(pb, level + 1, flags);
+		if (flags & JSON_C_TO_STRING_COLOR)
+			printbuf_strappend(pb,"\033[0;34m");
```
hawicz, line 478: Instead of having rather opaque escape sequences scattered around it would be better to add some defines and use them instead.  e.g. perhaps:
```
#define ANSI_COLOR_RESET "\e[0m"
#define ANSI_COLOR_FG_GREEN "\e[0;32m"
#define ANSI_COLOR_FG_BLUE "\e[0;34m"
#define ANSI_COLOR_FG_MAGENTA "\e[0;35m"
...
   printbuf_strappend(pb, ANSI_COLOR_FG_BLUE);
...
```
etc...

I'm not sure if there are commonly used names for these; if they exist, it would be a good idea to find and use those names.

hawicz:
There are a few things to tweak here, but I'm aiming to merge this for the upcoming 0.17 release.

cosmo-ray:
hello, thanks for the review, I've fix your comment.

Issue #705, opened 2021-05-04
segmentation fault on json-c parsing methods in cross compiled target


ausmauricio:
Hi, 

I'm trying to embed our software which uses json-c on a MIPS 32 target. Methods to build JSON such as `json_object_new_object()` work fine but parsing with for example `json_tokener_parse()` seg faults in this target. I think there is a problem with our cross-compilation because it works on other platforms, so any help here is much appreciated.

To simplify things I'm focusing on the app provided json_parse, which also seg faults, so I'm thinking if json_parse works so will our code. On the target:

> [root@AONT: home]# ./json_parse jsonsample  
> maxrss: 304 KB
> Segmentation fault (core dumped)

From what I understood running the same thing on my machine the output should be something like this:

> $ ./json_parse jsonsample 
> maxrss: 2756 KB
> Successfully parsed object from jsonsample
> [ jsonsample content here ]
> maxrss: 2756 KB
> maxrss: 2756 KB

file utility output on the target:

> $ file json_parse 
> json_parse: ELF 32-bit MSB executable, MIPS, MIPS32 version 1 (SYSV), dynamically linked, interpreter /lib/ld-uClibc.so.0, with debug_info, not stripped

I'm building for the target from the master specifying this toolchain file:

> root@e318088542a9:/usr/src# cat toolchainfile.cmake 
> set(CMAKE_SYSTEM_NAME Linux)
> set(CMAKE_C_COMPILER /opt/toolchains/crosstools-mips-gcc-4.6-linux-3.4-uclibc-0.9.32-binutils-2.21/usr/bin/mips-unknown-linux-uclibc-gcc)
> set(CMAKE_CXX_COMPILER /opt/toolchains/crosstools-mips-gcc-4.6-linux-3.4-uclibc-0.9.32-binutils-2.21/usr/bin/mips-unknown-linux-uclibc-g++)
> set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -lm -Os -pipe -Os  -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64" CACHE STRING "Buildroot CFLAGS" FORCE)
[...]

PR #777, opened 2022-07-04
Don't try to change locale when libc only supports the C locale

jameshilliard:
When uclibc is build only with C locale support we don't need to try and change to the C locale.

### DIFF
--- a/json_tokener.c
+++ b/json_tokener.c
@@ -298,11 +298,13 @@ struct json_object *json_tokener_parse_ex(struct json_tokener *tok, const char *
 	unsigned int nBytes = 0;
 	unsigned int *nBytesp = &nBytes;
 
+#ifndef __LOCALE_C_ONLY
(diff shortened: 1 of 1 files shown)


json_tokener.c
```diff
@@ -298,11 +298,13 @@ struct json_object *json_tokener_parse_ex(struct json_tokener *tok, const char *
 	unsigned int nBytes = 0;
 	unsigned int *nBytesp = &nBytes;
 
+#ifndef __LOCALE_C_ONLY
```
hawicz, line 301: Where does `__LOCALE_C_ONLY` get set?  I would expect to see something in CMakeLists.txt

hawicz:
It turns out that even with locale support available, uClibc's duplocale() is crashes (see issue #705) so I ended up changing CMakeLists.txt to entirely disable locale handling there.  With that in place, I believe this PR is unnecessary.

PR #779, opened 2022-07-12
Fix compiling for Android

m-kuhn:
Compiler complains

>  error: 'open' has superfluous mode bits; missing O_CREAT?


hawicz:
No, we definitely don't want to be creating a file here.

m-kuhn:
Good point and thanks for fixing!

Issue #780, opened 2022-07-24
Memory Leak when setting empty strings when c_string.pdata is used


dddaniel:
The current logic with below 0 size values indicating that a separate pdata buffer is used
breaks when setting an empty string in an existing json string object with json_object_string_set().

Reason is the pdata buffer is used, but the string length is zero. Zero means the current code will falsely assume no extra pdata buffer is used and therefore not free the old one.

https://github.com/json-c/json-c/blob/d28ac67dde77566f53a97f22b4ea7cb36afe6582/json_object.c#L1340-L1341

The issue can be easily reproduced with this little program running with valgrind --leak-check=full :

```
#include <json-c/json.h>

int main()
{
	json_object *obj = json_object_new_string("data");
	int i;

	for (i = 0; i < 4; i++) {
		/* Causes use of external malloced pdata buffer */
		json_object_set_string(obj, "Even longer data");

		/* Sets the string length to 0, causing the code to assume no pdata is used */
		json_object_set_string(obj, "");

		/* Sets longer data again, causing new pdata to be alloced
		 * while not free'ing the old one because the old length is zero and not negative */
		json_object_set_string(obj, "Even longer data");

		/* repeat the cycle */
		json_object_set_string(obj, "");
	}

	json_object_put(obj);
	return 0;
}
```

Valgrind output:
```
==3341818== Memcheck, a memory error detector
==3341818== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3341818== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==3341818== Command: ./test
==3341818== 
==3341818== 
==3341818== HEAP SUMMARY:
==3341818==     in use at exit: 136 bytes in 8 blocks
==3341818==   total heap usage: 9 allocs, 1 frees, 193 bytes allocated
==3341818== 
==3341818== 68 bytes in 4 blocks are definitely lost in loss record 1 of 2
==3341818==    at 0x483F7B5: malloc (vg_replace_malloc.c:381)
==3341818==    by 0x4879612: _json_object_set_string_len (json_object.c:1337)
==3341818==    by 0x4879707: json_object_set_string (json_object.c:1360)
==3341818==    by 0x109192: main (in /tmp/test)
==3341818== 
==3341818== 68 bytes in 4 blocks are definitely lost in loss record 2 of 2
==3341818==    at 0x483F7B5: malloc (vg_replace_malloc.c:381)
[...]

PR #781, opened 2022-07-24
Fix memory leak with emtpy strings in json_object_set_string

dddaniel:
When a json string object is updated with a bigger string, a new
malloc'ed buffer is used to store the new string and it's size is made
negative to indicate that an external buffer is in use.

When that same json string object get's updated again with an empty
stirng (size = 0), the new external malloc'ed buffer is still used.
But the fact that the new size value is not negative removes the
indicator that the externally malloc'ed buffer is used.

This becomes a problem when the object get's updated again with any
other string, because a new buffer will be malloced and linked to the
object while to old one won't be free'd.

This causes a memory leak when updating a json string with
json_object_set_stirng() which has previously been updated
with an empty string.

Example:
--
obj = json_object_new_string("data");
json_object_set_string(obj, "more data");
json_object_set_string(obj, "");
json_object_set_string(obj, "other data"); /* leaks */
--

This commit fixes the issue by free'ing the external buffer when an
empty string is set and use the internal one again.

Signed-off-by: Daniel Danzberger <daniel@dd-wrt.com>

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -1323,11 +1323,18 @@ static int _json_object_set_string_len(json_object *jso, const char *s, size_t l
 		// length as int, cap length at INT_MAX.
 		return 0;
 
-	dstbuf = get_string_component_mutable(jso);
 	curlen = JC_STRING(jso)->len;
-	if (curlen < 0)
-		curlen = -curlen;
+	if (curlen < 0) {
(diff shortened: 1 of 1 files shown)


dddaniel:
There is an issue for this PR #780

PR #789, opened 2022-09-08
Unnecessary struct declaration and unsafe function usage

EibzChan:
Struct 'json_object' forward declaration unnecessary, already declared. Please delete it.
According to standards for C Language, high Risk to  use strcat() in json_object.c . It should be replaced by strncat.
High Risk to  use vsprintf(). It should be replaced by vsnprintf.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -1047,7 +1047,7 @@ static int json_object_double_to_json_string_format(struct json_object *jso, str
 		{
 			// Ensure it looks like a float, even if snprintf didn't,
 			//  unless a custom format is set to omit the decimal.
-			strcat(buf, ".0");
+			strncat(buf, ".0", 2);
 			size += 2;
 		}
 		if (p && (flags & JSON_C_TO_STRING_NOZERO))
--- a/json_object_private.h
+++ b/json_object_private.h
@@ -20,7 +20,6 @@
 extern "C" {
 #endif
 
-struct json_object;
 #include "json_inttypes.h"
 #include "json_types.h"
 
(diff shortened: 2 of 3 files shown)


json_object.c
```diff
@@ -1047,7 +1047,7 @@ static int json_object_double_to_json_string_format(struct json_object *jso, str
 		{
 			// Ensure it looks like a float, even if snprintf didn't,
 			//  unless a custom format is set to omit the decimal.
-			strcat(buf, ".0");
+			strncat(buf, ".0", 2);
```
hawicz, line 1050: gcc 9 has a rather bogus warning on this line, but this is pointless change anyway and just serves to introduce potential discrepancies between the constant string and the length passed.

json_object_private.h
```diff
@@ -20,7 +20,6 @@
 extern "C" {
 #endif
 
-struct json_object;
```
hawicz, line 23: This is here to allow json_object_private.h to be included on its own.  If this is removed it's possible to result in a `warning: ‘struct json_object’ declared inside parameter list `

vasprintf_compat.h
```diff
@@ -51,7 +51,7 @@ static int vasprintf(char **buf, const char *fmt, va_list ap)
 		return -1;
 	}
 
-	if ((chars = vsprintf(b, fmt, ap)) < 0)
+	if ((chars = vsnprintf(b, ((size_t)chars + 1), fmt, ap)) < 0)
```
hawicz, line 54: The code just above this already uses vsprintf to calculate the right length and allocate the memory needed.  Passing a length just adds unnecessary checking.

PR #802, opened 2022-12-23
Problem: modern CMake warns about version 2.8

yrashk:
The warning is as follows:

```
Compatibility with CMake < 2.8.12 will be removed from a future version of CMake.
```

Solution: set it at 2.8.12


hawicz:
Sounds good to me.  I'll note that current openwrt (mentioned in the comment near the line you're changing) doesn't even have a cmake package, and expects people to cross-compile from a host system, so it's likely safe to bump this up even more, though exactly what version of cmake to pick as a minimum I don't know.

PR #804, opened 2022-12-23
Problem: cmake 3.25.1 warns about CMP0042 not being set

yrashk:
Solution: set it explictly to OLD behavior


BILLZIONS:
Thank you

PR #810, opened 2023-03-10
docs: update to Internet Standard reference

dijonkitchen:


hawicz:
If we're going to change the standard we're targeting then we also need to at least note what's different, if not actually implement anything new.
Do you happen to have a comparison between RFC7159 and RFC8259?

dijonkitchen:
This may say it best: https://www.tbray.org/ongoing/When/201x/2017/12/14/RFC-8259-STD-90

ASISBusiness (commented):
Working properly

PR #813, opened 2023-04-12
Build apps only in project itself.

ssrlive:


CMakeLists.txt
```diff
@@ -62,6 +62,9 @@ include(CPack)
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 
+set(LIB_JSON_C_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE INTERNAL "json-c library main path" FORCE)
```
hawicz, line 65: Aren't variables similar to this _already_ set, or do we have something wrong/missing with the files we install under lib/cmake/json-c ?

ssrlive: This exported global variable here is allowing another project that depends on json-c project to find the path of the json-c project without installing and compiling json-c.

hawicz: Do you have an example of where/how this would be used?
Most places that I can find that use json-c do so with the installed version, but I see that some (e.g. [gdal's cmake files](https://github.com/OSGeo/gdal/blob/master/cmake/helpers/CheckDependentLibraries.cmake#L43) are able to get the paths they want with `get_target_property` w/o needing json-c to set a variable itself.

ssrlive: Like this situation
```cmake
add_subdirectory(deps/json-c)

...

include_directories(${LIB_JSON_C_DIR})
include_directories(${LIB_JSON_C_BIN_DIR})
```

PR #825, opened 2023-09-02
Add support for WASI

fredldotme:
Disables shared libraries for WASI only and makes sure compat doesn't redefine builtins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,10 @@ include(CMakePackageConfigHelpers)
 option(BUILD_SHARED_LIBS  "Default to building shared libraries" ON)
 option(BUILD_STATIC_LIBS  "Default to building static libraries" ON)
 
+if (CMAKE_SYSTEM_NAME STREQUAL "WASI")
+    set(BUILD_SHARED_LIBS OFF)
+endif()
+
 if (BUILD_SHARED_LIBS)
     add_definitions(-D JSON_C_DLL)
 endif()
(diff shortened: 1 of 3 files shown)


fredldotme:
Sidenote: I haven't checked whether `json_c_snprintf` is actually used anywhere, actually. But I made it a possibly unused function due to the assumption that, just because it's not used in one source file might not mean much.

hawicz:
I haven't read more than the first few lines of the diff, but already I can tell you I'm not going to merge this.  Please don't litter the code with `ifdef __wasi__` lines. Instead, figure out why the regular symbol/feature detection isn't working, and fix that.

fredldotme:
Is at least the CMake part kosher enough or are there complaints about that approach too?

hawicz:
Sure, I guess that's ok, though it's not clear to me why you're disabling shared libs and tests, so a comment explaining that would be nice.

Also, is "WASI" in this case "Web Assembly System Interface"?  Please at least mention that full name in the commit message, and ideally add a "Building on WASI" section to README.md (assuming there are some special steps needed beyond the Building on Unix section)

Finally, the duplication in snprintf_compat.h is bugging me.  Can you try to consolidate with the MSC/MINGW case, perhaps with a `#define _vsnprintf vsnprintf` for the `__wasi__` case?
(does WASI really have the same lack-of-termination bug?  argh!)

fredldotme:
Sure can add some more commentary.

Yes WASI is "WebAssembly System Interface", and it has no concept of shared libraries at the moment, so everything must be statically linked together.

I'll take a look at the compat parts in a bit.

hawicz:
Closing due to lack of updates.  If you still want this, please address the problems mentioned earlier and re-open.

PR #828, opened 2023-09-06
Put arc4random check under DISABLE_EXTRA_LIBS=OFF in CMakeLists.txt.

robert-pang:
Per arc4number [doc](https://linux.die.net/man/3/arc4random), `arc4random` requires `libbsd` extra library. So putting the API's check under `DISABLE_EXTRA_LIBS=OFF` for consistency.


hawicz:
Only linux requires linking libbsd to get arc4random.  This change will break detection of it on other platforms, so I'm not merging this.

robert-pang:
Thank you for review. That makes sense. I just checked further and turns out `arc4random` has been [added](https://sourceware.org/git/?p=glibc.git;a=commit;h=6f4e0fcfa2d2b0915816a3a3a1d48b4763a7dee2) to glibc also.

PR #831, opened 2023-09-14
rename WIN32 to _WIN32

neheb:
The latter is the proper macro defined by Windows headers.

Fixes compilation under at least clang-cl which mandates function declarations.

As an aside, this issue is hidden under MinGW because unistd.h there includes io.h, which defines read, write, and so forth.

### DIFF
--- a/tests/test_util_file.c
+++ b/tests/test_util_file.c
@@ -2,11 +2,11 @@
 #undef NDEBUG
 #endif
 #include "strerror_override.h"
-#ifdef WIN32
+#ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN
 #include <io.h>
 #include <windows.h>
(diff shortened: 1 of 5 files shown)


neheb:
example failure: https://github.com/mesonbuild/wrapdb/actions/runs/6180647043/job/16777414263

gitartpiano:
Are there any maintainers working on this?

tests/test_util_file.c
```diff
@@ -2,11 +2,11 @@
 #undef NDEBUG
 #endif
 #include "strerror_override.h"
-#ifdef WIN32
+#ifdef _WIN32_
```
hawicz, line 5: This looks like it has an extra underscore

neheb: Fixed.

PR #835, opened 2023-09-22
fix error des of 'error_memory'

iasunsea:
3dad694 and 6dd8618 , the decs mast keep the same as the enum change the post

### DIFF
--- a/json_tokener.c
+++ b/json_tokener.c
@@ -116,8 +116,8 @@ static const char *json_tokener_errors[] = {
 	"invalid string sequence",
 	"expected comment",
 	"invalid utf-8 string",
-	"buffer size overflow",
-	"out of memory"
+	"out of memory",
+	"buffer size overflow"
 };
 /* clang-format on */
 


hawicz:
You're right that something is broken here, but I don't think that's quite the right fix.
Looks like I messed up the fix in 3dad6941, and rather than _actually_ putting `json_tokener_error_memory` at the end of the list I put it just before the end (because I interpreted `json_tokener_error_size` to mean "the size of the json_tokener_error enum, rather than an actual error).
Instead of rearranging the strings I'm going to fix the enum order.

hawicz:
Done in commit ad8b8afa7.  I should probably release a 0.17.1 really soon now.

ndvbrk:
@hawicz I'm finding your statement a year after. 
Please do release a new tag, I would look forward to it.
(I also suggest to consider reopening this issue until that happens, for visibility)

Background:
I am trying to upgrade from version 0.14 to 0.17, 
in the purpose of better handling OOM errors and printing error descriptions correctly, in a setup with limited memory.
Reading the changelog and commit history, its been incrementally resolved over many commits.
I am currently based in a very conservative organisation that expect a stable, formal release.
We could use a specific commit of the master branch, but prefer not to.

Thank you again for maintaining this library.

hawicz:
For now, I cherry picked that commit over to the json-c-0.17 branch.

hawicz:
Fyi, 0.18 is available now instead:
https://github.com/json-c/json-c/tree/json-c-0.18
https://s3.amazonaws.com/json-c_releases/releases/index.html

PR #847, opened 2023-12-23
Add support for Commodore Amiga

sacredbanana:
Made changes to get json-c compiling for the Commodore Amiga. It is used in my ChatGPT client for Commodore Amiga called AmigaGPT https://github.com/sacredbanana/AmigaGPT.

Compilation for Amiga is as simple as using a Docker container. See the readme for details.


CMakeLists.txt
```diff
@@ -112,7 +129,9 @@ check_include_file(sys/resource.h   HAVE_SYS_RESOURCE_H) # for getrusage
 check_include_file("dlfcn.h"        HAVE_DLFCN_H)
 check_include_file("endian.h"       HAVE_ENDIAN_H)
 check_include_file("limits.h"       HAVE_LIMITS_H)
-check_include_file("locale.h"       HAVE_LOCALE_H)
+if (NOT AMIGA)
+    check_include_file("locale.h"       HAVE_LOCALE_H)
```
hawicz, line 133: overall the changes look ok, but I'm wondering why do you add explicit guards around some of these checks?  Does amiga have these headers and symbols, but they don't work as expected?

sacredbanana: image

Amiga has these headers but a lot of the functions json-c uses from them are not available unless you are compiling a POSIX compliant app.

hawicz: Wow, that's pretty broken of those libraries, but ISTM that the right way to work around isn't to add "AMIGA" special checks, but instead check for the symbols in addition to the headers.  In fact, we already do so with e.g. the `HAVE_USELOCALE` define: https://github.com/json-c/json-c/blob/master/CMakeLists.txt#L175
Can you figure out why HAVE_USELOCALE ends up defined if the symbol isn't actually available?

sacredbanana: Thanks for your feedback. I'll look into this :)

In the meantime I have already released a compiled library at http://aminet.net/package/dev/lib/json-c

But for sure I will work to make the cmake build as clean as possible to be accepted into this repo :)

CMakeLists.txt
```diff
@@ -281,6 +301,9 @@ if ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_C_COMPILER_ID}" STREQUAL
 	if ("${DISABLE_WERROR}" STREQUAL "OFF")
 	    set(CMAKE_C_FLAGS           "${CMAKE_C_FLAGS} -Werror")
 	endif()
+    if (AMIGA_M68K)
+        set(CMAKE_C_FLAGS           "${CMAKE_C_FLAGS} -fbaserel")
```
sacredbanana, line 305: This is absolutely required for Motorola 68k architecture.

CMakeLists.txt
sacredbanana: Ok after much frustration I discovered that the current version of the C runtime libraries haven't actually implemented all the functions despite them being defined in the header files. At least this is the case for the Motorola 68k. The PowerPC libraries are more complete and don't require such checks.

CMakeLists.txt
hawicz: Ok, sure.  What is `-fbaserel` anyway?  I'm having trouble finding it in any gcc docs.

CMakeLists.txt
hawicz: hmm... I wonder if we're running into issues here because you're cross-compiling, and cmake isn't using the right toolchain for the check_symbol_exist steps.  I notice that your readme instructions don't mention passing a `--toolchain` option as suggested at https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling
Can you perhaps try that and see if it improves the situation?

sacredbanana: The correct toolchain automatically gets used because its set in an environment variable baked in the Amiga gcc Docker images

CMakeLists.txt
sacredbanana: For Motorola 68k Amiga code, it will not successfully compile without that flag.

ChatGPT can explain it better than I can.

image

image

sacredbanana:
Bump. Are you fine with this change and want it merged in?

CMakeLists.txt
```diff
@@ -73,12 +69,47 @@ option(DISABLE_JSON_PATCH             "Disable JSON patch (RFC6902) support."
...
+endif()
 
-if (UNIX OR MINGW OR CYGWIN)
+if (${CMAKE_SYSTEM_NAME} STREQUAL "AmigaOS")
+    set(AMIGA ON)
+    set(DISABLE_THREAD_LOCAL_STORAGE ON)
+    set(ENABLE_THREADING OFF)
+    set(BUILD_SHARED_LIBS OFF)
+    set(BUILD_APPS OFF)
     list(APPEND CMAKE_REQUIRED_DEFINITIONS -D_GNU_SOURCE)
```
hawicz, line 82: This line is duplicating the same `-D_GNU_SOURCE` from a few lines above, so this should be removed from this block, but the block of code that sets if (lines 72-74) need to move a bit lower (so the check for `AMIGA` works)

CMakeLists.txt
```diff
@@ -87,7 +118,7 @@ if (MSVC)
     list(APPEND CMAKE_REQUIRED_FLAGS /wd4996)
 endif()
 
-if (NOT DISABLE_STATIC_FPIC)
+if (NOT DISABLE_STATIC_FPIC AND NOT AMIGA)
```
hawicz, line 121: If `CMAKE_POSITION_INDEPENDENT_CODE` doesn't work, how about setting `DISABLE_STATIC_FPIC` above, where you disable various other things?

CMakeLists.txt
hawicz: Looking at the docs for check_symbol_exists, it says `If the header files declare the symbol as a function or variable then the symbol must also be available for linking`

Is `uselocale` defined as a macro in your toolchain header files?  If not, then I would expect check_symbol_exists to be able to figure out that the symbol doesn't actually exist in the library.
If you add something like this, does cmake think the symbol exists in the library?
```
check_function_exists(uselocale HAVE_USELOCALE_IN_LIBRARY)
if (HAVE_USELOCALE_IN_LIBRARY)
message(STATUS "uselocale() IS present in library")
else()
message(STATUS "uselocale() is NOT present in library")
endif()
```

I think we need to figure out why the check isn't detecting the lack of the symbols in the actual library.  Then, if it turns out we do need some kind of exception, it should be down after `HAVE_USELOCAL`/`HAVE_SETLOCALE` are set, similar to the special case for uClibc (probably using the check_function_exists macro)

sacredbanana: OK so I removed the Amiga specific guard on locale.h because I figured out why it was broken when you want to build the clib2 version of json-c. Turns out there was a bug in clib2 where you will get an error about conflicting types if your project has an `#include` for both `locale.h` and `math.h` and you're building from Bebbo's Amiga dev toolchain. I have created a PR https://github.com/adtools/clib2/pull/18 to fix that issue. Now there are no locale issues when building json-c. Basically clib2 just didn't properly cater for other c runtime libraries in the environment other than itself. Also it seems newlib has been updated to add definitions for `use locale()`, `freelocale()` and `duplocale()` so those linking errors in the screenshot I posted a few months ago are no longer occurring.

CMakeLists.txt
sacredbanana: Good catch. Silly me

CMakeLists.txt
sacredbanana: Done

sacredbanana:
Changes made. Please review it when you're free thanks :)

CMakeLists.txt
hawicz: Awesome work figuring this out, thank you!
I'll review and merge this in the next week or so.

sacredbanana:
Turns out libnix doesn't support duplocale() despite the symbol existing. Added a check to disable it if building for libnix.

CMakeLists.txt
```diff
@@ -379,7 +385,7 @@ if (NOT ("${CMAKE_C_COMPILER_FRONTEND_VARIANT}" STREQUAL "MSVC"))
 
 	# OSX Mach-O doesn't support linking with '-Bsymbolic-functions'.
 	# Others may not support it, too.
-	list(APPEND CMAKE_REQUIRED_LIBRARIES "-Wl,-Bsymbolic-functions")
```
hawicz, line 347: Unconditionally adding `-lamiga` isn't going to work on most systems.
Adding `-lc` at all sounds like the wrong thing to do too.
If these are actually needed then it needs to be limited to just the environments where that's the case.  Also, it definitely shouldn't be lumped in with passing the `-Bsymbolic-functions` flag.

hawicz: How about adding a `list(APPEND CMAKE_REQUIRED_LIBRARIES "-Wl,-lamiga")` up in the `if (${CMAKE_SYSTEM_NAME} STREQUAL "AmigaOS")` block above?

sacredbanana: Oops this was added as a test and turns out it's not required at all. I've removed it.

sacredbanana:
It has been a while since this PR has been looked at. I did just add one more thing to it now which is adding support for the Amiga based OS called MorphOS

sacredbanana:
@hawicz bump. does this PR look ok? I'm keen to get this merged

sacredbanana:
Just fixed the merge conflict. Can I get this one reviewed please?

PR #864, opened 2024-06-17
Build pkg-config for msvc as well

pavelzw:
Hi, would it be possible to build pkg-config for MSVC targets as well?


hawicz:
Sure, I suppose, but when is this actually useful?  I didn't think there was a good way to use pkg-config on Windows.

Also, why drop the whitespace?  I think it's more readable when the stages of creating the json-c.pc file are separated a bit.

pavelzw:
When compiling gnucobol for windows, I found that I was missing `json-c.pc` 🤔 when manually adding it, the compilation worked.
I'll revert the whitespace changes 👍🏻

PR #878, opened 2024-11-03
Fix underline bug in README

0xfadead:
Under the "Usage"-section some text is erroneously underlined, because some `a`-tags aren't closed. All this pull-request does is close them.


j0k3r (approved):
I was about to submit the same PR :)

PR #879, opened 2024-11-06
Handle NULL gracefully in json_tokener_free

janotomko:
Similarly to glibc's free, make json_tokener_free(NULL) a no-op, to simplify cleanup paths.

### DIFF
--- a/json_tokener.c
+++ b/json_tokener.c
@@ -182,6 +182,8 @@ struct json_tokener *json_tokener_new(void)
 
 void json_tokener_free(struct json_tokener *tok)
 {
+	if (!tok)
+		return;
 	json_tokener_reset(tok);
 	if (tok->pb)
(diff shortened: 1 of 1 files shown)


hawicz:
I don't quite agree with the claim that this "reduces bugs".  "Reduces crashes" perhaps, but depending on doing useful stuff with NULL values is IMO sloppy.  Nevertheless, I can see the convenience of this change, and an extra check in the free function is unlikely to cause much performance impact so I'll merge this.

rwmjones:
It lets you use `__attribute__((cleanup))` with less code, which reduces memory leaks, and lets you unify exit paths from functions, which reduces code complexity and therefore bugs.  It's also consistent with `free` from the C standard and many other free-like functions, thus reducing surprise.

PR #882, opened 2024-11-20
Explicitly handle NaN values when converting to integer

simonresch:
Json objects of type double with the value NaN could cause undefined behavior when casting double to int in `json_object_get_int`.

This PR explicitly handles this case in the functions `json_object_get_int`, `json_object_get_in64` and `json_object_get_uint64`. Instead of the integer conversion of NaN which is implementation specific INT32_MIN/INT64_MIN/0 is returned and the error code set to `EINVAL`.


json_object.c
```diff
@@ -767,6 +767,11 @@ int32_t json_object_get_int(const struct json_object *jso)
 			return INT32_MIN;
 		if (cdouble >= INT32_MAX)
 			return INT32_MAX;
+		if (isnan(cdouble))
+		{
+			errno = EINVAL;
+			return 0;
```
hawicz, line 773: 0 seems to likely to be a "good" value, how about `INT32_MIN` instead?
Also, if we're setting errno on "errors", I think we need to first initialize it to 0 on entry to json_object_get_int()

To make this more complete, it would seem sensible to set `errno = ERANGE` in the various cases where we cap the value returned, but that's probably best as a separate commit.

simonresch: > 0 seems to likely to be a "good" value, how about INT32_MIN instead?
I went with 0 since that is returned for other conversion errors. But I think you're right that for code that does not handle errors when calling `json_object_get_int` it would be better to fail more visibly with INT32_MIN. That should also be less of a breaking change for the architectures x86 and x64.
> Also, if we're setting errno on "errors", I think we need to first initialize it to 0 on entry to json_object_get_int()
Good point. :+1: Done. 

> To make this more complete, it would seem sensible to set errno = ERANGE in the various cases where we cap the value returned, but that's probably best as a separate commit.
Sounds useful to me. I added a second commit to set the `ERANGE` in those cases.

If you think setting `errno` for these lossy conversions is "to much" I'm also happy to revert the error handling. In my opinion the important point part is that `json_object_get_int()` returns the same values e.g. on ARM and x64.

json_object.h
```diff
@@ -783,8 +786,12 @@ JSON_EXPORT int json_object_int_inc(struct json_object *obj, int64_t val);
 /** Get the int value of a json_object
  *
  * The type is coerced to a int64 if the passed object is not a int64.
- * double objects will return their int64 conversion. Strings will be
- * parsed as an int64. If no conversion exists then 0 is returned.
+ * double objects will return their int64 conversion except for NaN values
+ * which return 0 and the errno is set to EINVAL.
```
hawicz, line 790: the doc comments need to be adjusted too

simonresch: Ah, yes of course. Thanks!

PR #883, opened 2024-11-20
Add fuzz tests for json_object/point/array apis

simonresch:
This PR adds 3 fuzz tests for continuous fuzz testing in [OSS-Fuzz](https://github.com/google/oss-fuzz). The fuzz tests include usages of `json_object_array_*`, `json_object_get_*`, `json_object_set_*`, `json_pointer_*` API functions which are not covered in the existing `tokener_parse_ex_fuzzer`.

A local trial run with the OSS-Fuzz introspector reported a combined coverage of 78% compared to 42% [currently reported in OSS-Fuzz](https://introspector.oss-fuzz.com/project-profile?project=json-c).

Note, for the fuzz tests in this repository to be automatically executed https://github.com/google/oss-fuzz/pull/12741 also needs to be merged.


fuzz/json_array_fuzzer.dict
```diff
@@ -0,0 +1,19 @@
...
+"e-"
+"E"
+"E+"
+"E-"
+"\""
+"null"
+"1"
+"1.234"
+"3e4"
+"NaN"
```
hawicz, line 19: Would it be useful to include `Infinity`/`-Infinity` in the dictionaries?

simonresch: Yes, good point. Added.

PR #890, opened 2025-04-15
random_seed.c: get_getrandom_seed(): use GRND_INSECURE if available

jacmet:
As explained in https://github.com/json-c/json-c/pull/832, getrandom(.., GRND_NONBLOCK) fails with EAGAIN if the Linux kernel random number pool is not yet initialized.

The use case here is initializing the hash table seed rather than strong randomness for crypto, so use GRND_INSECURE if available (Linux 5.6+) to make getrandom() return best effort random data during bootup rather than fail.

### DIFF
--- a/random_seed.c
+++ b/random_seed.c
@@ -176,6 +176,12 @@ retry:
 #include <sys/random.h>
 #endif
 
+/* Return best effort random data even if random pool is not yet
+ * initialized. Available since Linux 5.6 */
+#ifndef GRND_INSECURE
+#define GRND_INSECURE 0
+#endif
+
 static int get_getrandom_seed(int *seed)
 {
 	DEBUG_SEED("get_getrandom_seed");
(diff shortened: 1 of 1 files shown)


hmh (approved):

jacmet:
@hmh what is needed to get this merged?

hmh:
> @hmh what is needed to get this merged?

Ask @hawicz... I just looked over the patch as a third party eye on something that could be relevant for my own uses of json-c.

IMHO, even if extra changes are desired at a higher level (e.g. to have an userspace alternative PRNG to ark4random that can be selected at build-time when one knows ark4random would seed in blocking mode), the proposed change to get_rdrand_seed() is desirable.

jacmet:
@hawicz ping?

jacmet:
@hawicz ping?

PR #893, opened 2025-07-31
Fix bug involving supplemental code points that look like high surrogates

sffc:
Hi there! I came across a bug when running Unicode conformance tests on a platform that uses `json-c` and eventually traced it to here.

In the main parse loop, the code reads a surrogate pair encoded as `\uXXXX\uYYYY`, and then it decodes it as follows:

```c
					tok->ucs_char = DECODE_SURROGATE_PAIR(tok->high_surrogate,
					                                      tok->ucs_char);
```

This is fine so far. But then back in the main loop:

```c
			if (tok->ucs_char < 0x80)
			{
				// handle 1-byte character
			}
			else if (tok->ucs_char < 0x800)
			{
				// handle 2-byte character
			}
			else if (IS_HIGH_SURROGATE(tok->ucs_char))
			{
				// handle high surrogate
			}
			else if (IS_LOW_SURROGATE(tok->ucs_char))
			{
				// handle unpaired low surrogate
			}
			else if (tok->ucs_char < 0x10000)
			{
				// handle 3-byte character
			}
			else if (tok->ucs_char < 0x110000)
			{
				// handle 4-byte character
			}
			else
			{
				// unknown error
			}
```

At this point in the parse, we should be hitting the `if (tok->ucs_char < 0x110000)` case, and most of the time we do. However, for some supplemental code points, we are hitting the `if (IS_HIGH_SURROGATE(tok->ucs_char))` case. The definition of `IS_HIGH_SURROGATE` is:

```c
#define IS_HIGH_SURROGATE(uc) (((uc)&0xFC00) == 0xD800)
```

This correctly identifies BMP code points that are high surrogates. However, characters in multiple supplemental blocks also pass this test, including [SignWriting](https://en.wikipedia.org/wiki/Sutton_SignWriting_(Unicode_block)) (code points in range U+1D800..U+1DBFF) and [CJK Unified Ideographs Extension F](https://en.wikipedia.org/wiki/CJK_Unified_Ideographs_Extension_F) (code points in range U+2D800..U+2DBFF), and also various private-use ranges.

To fix this, I changed the macro to mask all bits in the 32-bit integer, instead of just the bottom 16 bits of the 32-bit integer.

CC @sven-oly @echeran

### DIFF
--- a/json_tokener.c
+++ b/json_tokener.c
@@ -145,8 +145,8 @@ enum json_tokener_error json_tokener_get_error(struct json_tokener *tok)
 }
 
 /* Stuff for decoding unicode sequences */
-#define IS_HIGH_SURROGATE(uc) (((uc)&0xFC00) == 0xD800)
-#define IS_LOW_SURROGATE(uc) (((uc)&0xFC00) == 0xDC00)
+#define IS_HIGH_SURROGATE(uc) (((uc)&0xFFFFFC00) == 0xD800)
+#define IS_LOW_SURROGATE(uc) (((uc)&0xFFFFFC00) == 0xDC00)
 #define DECODE_SURROGATE_PAIR(hi, lo) ((((hi)&0x3FF) << 10) + ((lo)&0x3FF) + 0x10000)
 static unsigned char utf8_replacement_char[3] = {0xEF, 0xBF, 0xBD};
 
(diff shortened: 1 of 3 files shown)


sffc:
Hi @hawicz, can you advise on the next steps here? Thank you!

hawicz:
lgtm, merging

PR #899, opened 2025-10-20
Meson support

vonericsen:
Adding meson build system support.

When I was debugging the issue in the Cmake build for clang-cl (#898) I also implemented a meson build system that can build json-c using meson as well.
I implemented this since I understood meson much better than Cmake and wanted to ensure it was not an issue specific to Cmake.
Once I got this working in meson, I went back and fixed the cmake build.

I was able to run the tests in Linux and Windows. I had to modify the test_util_file.c for Windows though since it was causing a crash when opening `/dev/null` failed, which was a simple return statement after the error.
I tried to make this use the special file `NUL` for Windows instead, but the test will always fail with a CRT crash/timeout since this test is trying to make sure opening an invalid file descriptor fails, but windows will always crash from what I saw (in debug and release modes).
The test also had some issues reading from the files and having incorrect sizes. In Windows O_BINARY needs to be used in the flags due to how it handle line endings and translating them. While I fixed that in this test, I am not sure whether this needs fixes elsewhere within json-c.

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 2.8)  # see ../CMakeLists.txt for why 2.8
+cmake_minimum_required(VERSION 3.9...3.12)  # see ../CMakeLists.txt for why
 
 if(POLICY CMP0075)
     cmake_policy(SET CMP0075 NEW)


hawicz:
I'll merge this, but I don't know anything about meson, so it'll be up to you to keep that up-to-date.
Before I do, please rebase since I merged the "Fixing Cmake build..." change in a separate MR.

PR #903, opened 2025-11-23
meson: use override_dependency

neheb:
Allows wraps to use the dependency_names syntax instead of the possibly changing variable syntax.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -1,0 +1,228 @@
+name: meson
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{github.workflow}}-${{github.head_ref}}
+  cancel-in-progress: true
+
+jobs:
+  Linux-GCC:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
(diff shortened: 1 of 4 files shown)


vonericsen (commented):
The meson config changes look good to me.

The changes for strcasecmp also look more correct than my original versions since I had not tested mingw previously, but checking _WIN32 is more correct from my experience on windows with various compilers.

The changes for dllimport/dllexport vs def file I think there are considerations for the maintainer to make. These do look correct though from reading through the changes.

neheb:
Yeah. On Windows def files are much simpler as otherwise both dllexport and dllimport are needed. Only the former is present currently.

json-c.def
```diff
@@ -0,0 +1,150 @@
+EXPORTS
+    ; JSONC_PRIVATE
+    array_list_add
```
hawicz: These symbols should not be listed here.  These are not marked with JSON_EXPORT, and so would not actually be exported on Windows builds.

neheb: Yep. Currently they are exported on non Windows builds. I’ll work around this for meson.

Issue #916, opened 2026-01-25
OOM via large array index in json_pointer_set


thely314:
**Describe the bug**
An OOM vulnerability exists in the `json_pointer_set` function (and related functions) due to uncontrolled resource consumption. When handling a JSON Pointer path that targets an array with an extremely large numerical index (e.g., `"/4444444444"`), the library attempts to allocate a massive amount of memory proportional to this index. This leads to an immediate OOM condition, causing the OS's OOM Killer to terminate the process.

An attacker can exploit this by sending a very small JSON object (e.g., `[]`) and a specially crafted JSON Pointer path, making it an effective DoS vector.

The root cause lies in the `is_valid_index` function (`json-c/json_pointer.c line 46`), which uses `strtoull` to convert the path segment to a `size_t` index but fails to perform a sanity check on the magnitude of the resulting value. This large, but valid, `size_t` value is then passed down to `array_list_expand_internal` (`json-c/arraylist.c line 81`), which triggers an enormous `realloc` request.

**Steps To Reproduce**
The vulnerability can be reproduced with the following minimal cpp code, which was derived from fuzzing with LLVM's libFuzzer.

1. minimal cpp code
```cpp
// reproducer.cpp
#include <iostream>
#include <string>

#include <json_object.h>
#include <json_tokener.h>
#include <json_pointer.h>

int main(int argc, char **argv) {
    // The attack vector: a simple array and a JSON Pointer path with a huge index.
    std::string json_str = "[]";
    std::string path_str = "/4444444444";

    json_object* root = json_tokener_parse(json_str.c_str());
    if (!root) {
        std::cout << "Failed to parse JSON string. The input might be invalid JSON." << std::endl;
        root = json_object_new_object();
    }
    std::cout << "JSON object created." << std::endl;

    std::cout << "Attempting to call json_pointer_set with path: " << path_str << std::endl;
    json_object* value_to_set = json_object_new_int(123);

    // This call will trigger a massive memory allocation request and be terminated by the OOM Killer.
    int result = json_pointer_set(&root, path_str.c_str(), value_to_set);

    // The program will likely not reach this point.
    if (result == 0) {
        std::cout << "json_pointer_set succeeded." << std::endl;
    } else {
        std::cout << "json_pointer_set failed with code: " << result << std::endl;
        json_object_put(value_to_set);
    }
    json_object_put(root);
    return 0;
}
```

2. compile & run the code
```sh
# assuming you are currently outside the json-c directory
mkdir -p build
[...]

PR #917, opened 2026-01-25
Fix: OOM vulnerability cause by is_valid_index

thely314:
Hi, An OOM vulnerability exists in the `json_pointer_set` function (and related functions). See issue #916 for more details.

To fix that, added a sanity check in the `is_valid_index` function to limit the maximum value of a parsed array index. Also provided a configurable macro for modification.

### DIFF
--- a/json_pointer.c
+++ b/json_pointer.c
@@ -423,3 +423,39 @@ out:
 	free(path_copy);
 	return rc;
 }
+
+int json_pointer_set_with_limit_index(struct json_object **obj, const char *path,
+                                    struct json_object *value, size_t limit_index)
+{
+	// -1 means no limits
+    if (limit_index == (size_t)-1)
+    {
+        return json_pointer_set_with_array_cb(obj, path, value, json_object_array_put_idx_cb, NULL);
+    }
+    return json_pointer_set_with_array_cb(obj, path, value,
+                      json_object_array_put_idx_with_limit_cb, &limit_index);
+}
+
+/* safe callback for array index limit */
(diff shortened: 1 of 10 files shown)


json_pointer.c
```diff
@@ -413,3 +413,40 @@ int json_pointer_setf(struct json_object **obj, struct json_object *value, const
...
+
+/* safe callback for array index limit */
+int json_object_array_put_idx_with_limit_cb(struct json_object *jso, size_t idx,
+                                       struct json_object *jso_new, void *priv)
+{
+	size_t max_idx;
+    // use priv as a size_t pointer to pass in the maximum allowed index
+	if (!priv)
+	{
+		errno = EINVAL;
```
hawicz: This seems more like a programming error that I'd be inclined to just assert() on.  At the very least it should return something different from the "too big" case, perhaps `EFAULT`

hawicz:
Generally these changes look ok.  I'm mulling over whether it would be worthwhile to have a more general callback that applies to setting fields in objects too.

thely314:
I've made the callback more generic, which involved modifications to the original callback logic. Please review these changes carefully as they may require further discussion.

Also added tests for the three newly exported public functions . All tests (including existings) passed in my docker env.
My fuzzer ran for 2h with no more issues discovered.

thely314:
Hi @hawicz , just wanted to check if you’ve had a chance to look at this PR when you have a moment. All tests are passing, appreciate your time!

hawicz:
looks good, thanks for the patch

PR #925, opened 2026-05-12
Fix build with clang-21

mbroz:
This patch fixes the build error with clang-21
  tests/test_deep_nesting.c:19:34: error: a function declaration
  without a prototype is deprecated in all versions of C
  [-Werror,-Wstrict-prototypes]
   19 | static char *generate_json_string()

by fixing generate_json_string() definition.

It also removes prototypes for static functions that are not needed.

### DIFF
--- a/tests/test_deep_nesting.c
+++ b/tests/test_deep_nesting.c
@@ -15,8 +15,7 @@
 
 #define NESTING_DEPTH 100000
 
-static char *generate_json_string(void);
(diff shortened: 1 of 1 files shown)


mbroz:
@hawicz Could you please merge this (or fix the build a different way)? 

We are using json-c master branch in cryptsetup oss-fuzz CI build and it is broken for more than two weeks already due to this trivial issue.
Thanks.

PR #930, opened 2026-06-15
fix locale-dependent strtod in json_object_get_double

dxbjavid:
json_object_get_double() parses a json_type_string with strtod(), which honours the current LC_NUMERIC locale, even though the value is stored with a '.' radix the way it appears in JSON. Under a locale whose decimal point is ',' (for example de_DE) the '.' is then treated as a stray trailing character, so a string value like "19.95" comes back as 0 rather than the number it represents. This parses a copy with the radix translated to the locale's decimal point, so the result no longer depends on LC_NUMERIC, which lines up with the locale-independent number handling the tokener already does.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -21,6 +21,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef HAVE_LOCALE_H
+#include <locale.h>
+#endif /* HAVE_LOCALE_H */
 
 #include "arraylist.h"
 #include "debug.h"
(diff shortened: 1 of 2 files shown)


json_object.c
```diff
@@ -1381,12 +1384,42 @@ double json_object_get_double(const struct json_object *jso)
...
+		const char *parse = cstr;
+		char *radixconv = NULL;
+#ifdef HAVE_LOCALE_H
+		/* A json_type_string holds the value with a '.' radix, as it
+		 * appears in JSON, but strtod() honours the current locale.  In a
+		 * locale whose decimal point is not '.' the '.' is treated as a
+		 * stray character, so e.g. "19.95" is read as 0.  Parse a copy
+		 * with the radix translated so the value comes back unchanged. */
+		const char *decimal_point = localeconv()->decimal_point;
+		if (decimal_point && decimal_point[0] != '.' && decimal_point[1] == '\0')
```
hawicz, line 1398: It's unfortunate we don't have a way for the caller of the json-c API to tell us ahead of time whether they'll ensure that the locale is set to a compatible one.  Lacking that, I don't see a way to avoid the overhead of calling localconv() every time through here.  Oh well.

dxbjavid:
Thanks for the merge.

PR #931, opened 2026-06-24
deep copy values in json_patch copy op to avoid aliasing and cycles

dxbjavid:
the json patch copy operation shares the source object by reference rather than duplicating it, so the value at the new location is the same object as the source. besides changes to the copy being visible in the original, a later copy whose from resolves to that shared object can insert it beneath itself through a different pointer path, which the textual from/path check in json_patch_apply_move_copy cannot detect, leaving a reference cycle that recurses until the stack overflows when the result is serialised (reachable from any untrusted patch document). copy now deep copies the value as RFC 6902 section 4.5 requires, so the new subtree is independent and no cycle can form; move is unchanged and still transfers the original object. a json_type_null source is inserted directly since it cannot form a cycle, and a regression case is added to tests/json_patch_tests.json.

### DIFF
--- a/json_patch.c
+++ b/json_patch.c
@@ -198,6 +198,7 @@ static int json_patch_apply_move_copy(struct json_object **res,
 	json_pointer_set_cb array_set_cb;
 	struct json_pointer_get_result from;
 	struct json_object *jfrom;
+	struct json_object *value;
 	const char *from_s;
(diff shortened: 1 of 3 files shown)


dxbjavid:
Thank you for reviewing and merging this PR.

PR #933, opened 2026-06-27
fix signed overflow negating INT64_MIN in json_object_int_inc

dxbjavid:
json_object_int_inc works out the magnitude of a negative increment with -val, but that negation is undefined when val is INT64_MIN, since the result does not fit back into an int64_t. On the uint64 cint_type branch the expression is used in both range comparisons and the subtraction, so incrementing a uint64-typed value by INT64_MIN runs into signed overflow and UBSan aborts there. Doing the negation in unsigned space with 0 - (uint64_t)val keeps the magnitude well defined for the whole int64_t range and leaves the result unchanged for every other value. I have added a case to test_int_add covering the INT64_MIN decrement on a uint64 object.


hawicz:
Thanks for the PR!

dxbjavid:
Thank you for the review and merge. Much Appreciated.

PR #935, opened 2026-06-29
Nullify _userdata after _user_delete callback

arkadesOrg:
Nullify _userdata after callback _user_delete is applied.

```c
if (jso->_user_delete) {
	jso->_user_delete(jso, jso->_userdata);
	jso->_userdata = NULL;
}
```

This prevents _json_object_put_maybe_free() during a final assert of json_object_put() to pass a danglin pointer. 
```c
assert(_json_object_put_maybe_free(jso, 1) == 0);
```
Maybe its even better to hand over more control to the user by passing userdata by reference [as suggested here](https://github.com/json-c/json-c/pull/937):

```c
typedef void(json_object_delete_fn)(struct json_object *jso, void **userdata);
```

This addresses the [issue 934](https://github.com/json-c/json-c/issues/934)

PR #936, opened 2026-06-30
free objects outside assert() in json_object_put

dxbjavid:
json_object_put frees each non-empty container on the way back up its iterative teardown by calling _json_object_put_maybe_free(jso, 1) inside an assert(). Under NDEBUG, which is the default for release and most distro builds, assert() expands to nothing and the call is never evaluated, so every object/array node in the tree (along with its lh_table/array_list) leaks. I confirmed it with the leaks tool on a small nested document: 16 allocations leaked when built with -DNDEBUG, none with asserts on. The test suite builds with asserts enabled so it doesn't show up there. Evaluate the call into a local first and assert on that instead, so the free always happens.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -435,11 +435,16 @@ int json_object_put(struct json_object *jso)
 		// All slots are cleared, now pop back up to the parent
 		{
 			json_object *parent = jso->_delete_parent;
+			int rc;
 			// jso is a child that's already been detached from its parent
 			// so we need to actually free it now
 			assert(jso->_ref_count == 0);
 			jso->_ref_count++;   // We're the exclusive owner of jso, non-atomic add is ok.
-			assert(_json_object_put_maybe_free(jso, 1) == 0);
+			// Note: the call must not be inside assert(), or it gets
+			// compiled out when NDEBUG is defined and the memory leaks.
+			rc = _json_object_put_maybe_free(jso, 1);
+			assert(rc == 0);
+			(void)rc;
 			jso = parent;
 			// iteration will be reset at the top of the loop
 		}


arkadesOrg:
the issue is even worse as the assertion can [cause passing a danglin pointer to _user_delete](https://github.com/json-c/json-c/pull/935)

dxbjavid:
agreed that's worth fixing, though i think it's a separate concern from this one. the _userdata handling in #935 sits in the _user_delete path, whereas this change only affects whether the free runs at all: under NDEBUG the assert compiles the call out entirely, so the node and its table leak. the call already runs in debug builds today, so moving it out of the assert doesn't alter the _user_delete behaviour either way, it just makes teardown consistent in release builds. the two look complementary to me.

arkadesOrg:
you are right, it is basically a separate concern, even though its related to the same path.

currently I cannot build against latest json-c version, as it causes lots of crashes with _userdata containing objects.

I'd appreciate both fixes a lot.

dxbjavid:
appreciate you confirming. the change here is about as small as it gets and the leak is gone under NDEBUG in my testing, so it should be safe to take on its own. for what it's worth i triggered it with a small nested document under the leaks tool, 16 allocations leaked with -DNDEBUG and none with asserts on, happy to attach that if a maintainer wants a reproducer. the existing suite builds with asserts enabled so it can't really catch this one, which is why it slipped through. hopefully both can land.

arkadesOrg:
I am wondering how the tests are conducted, because the dangling pointer should cause a double free of _userdata with asserts on.

dxbjavid:
i had a look at that. the short answer is the suite never really exercises a deleter on a nested container, so there's nothing to trip. the only test that sets one is test_set_serializer, and it deliberately doesn't free anything (the userdata is stack allocated), plus it's a flat object that tears down through the single-object path rather than the iterative one where _delete_parent aliases _userdata. so even with asserts on, no test frees _userdata twice.

and to be clear that double free is the #935 path, not this one. the leak i'm fixing here only shows up under NDEBUG, where the free never runs at all, so the two don't overlap. a small test that puts a deleter on a container inside a larger tree would be the thing that catches the #935 case, but that's really for that PR to add.

hawicz:
Thanks for the fix!

dxbjavid:
TY!

PR #937, opened 2026-06-30
userdata by reference

arkadesOrg:
Pass the _userdata to _user_delete by reference to give user more control over _userdata nullification.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -302,11 +302,9 @@ static inline int _json_object_put_maybe_free(struct json_object *jso, int free_
 	}
 
 	if (jso->_user_delete) {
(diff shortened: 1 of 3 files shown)


hawicz:
Sorry, I can't merge this.  For one, this is a major API change and would break any existing user_delete functions.
Also, because `_user_delete()` is called just before a jso is freed there's no point in giving `_user_delete()` any choice about whether to clear `jso->_userdata` because it's going to be rendered invalid momentarily anyway, when the `struct json_object *` is freed.

arkadesOrg:
right, the radical API change is insane as `struct jso_object *` is free'd anyway 🕵️

PR #939, opened 2026-07-02
tests: do away with xxd invocation

jengelh:
printf is perfectly capable of doing everything needed.


hawicz:
Only the `printf` from bash handles escape sequences like this, other shells do not.  In particular, `/bin/dash`, which is what `/bin/sh` is often a symlink to, does not.

jengelh:
So let's use /usr/bin/printf then?

PR #940, opened 2026-07-06
reject out-of-range double in json_object_get_int64/get_uint64

dxbjavid:
json_object_get_int64() and json_object_get_uint64() guard the double-to-integer cast with c_double > (double)INT64_MAX (and the UINT64_MAX equivalent), but neither INT64_MAX nor UINT64_MAX is representable as a double, so the cast rounds the bound up to 2^63 and 2^64 respectively. A double holding exactly that rounded value slips past the strict > comparison and is then cast to int64_t/uint64_t, which is out of range and undefined behaviour; a json_type_double of 9223372036854775808.0 or 18446744073709551616.0 (reachable straight from json_tokener_parse) makes UBSan abort at the cast. Switching the two checks to >= clamps that boundary value to the max and sets ERANGE, matching how the existing infinity and string paths already report out-of-range, and leaves every in-range value untouched.

### DIFF
--- a/json_object.c
+++ b/json_object.c
@@ -996,9 +996,10 @@ int64_t json_object_get_int64(const struct json_object *jso)
 		}
 	}
 	case json_type_double:
-		// INT64_MAX can't be exactly represented as a double
-		// so cast to tell the compiler it's ok to round up.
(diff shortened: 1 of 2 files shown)


dxbjavid:
TY!

PR #942, opened 2026-07-13
fix out-of-bounds read for negative errno in _json_c_strerror

dxbjavid:
the numeric fallback in _json_c_strerror() formats an unknown errno by indexing "0123456789" with errno_in % 10, but for a negative errno_in that modulo is negative and reads before the start of the literal, which addresssanitizer reports as a global-buffer-overflow. the override path is reachable through the exported symbol when _JSON_C_STRERROR_ENABLE is set (as the tests do), and negating INT_MIN to get the magnitude would overflow as well. i've reworked the conversion to build the value in unsigned space and prepend a minus sign, so the output for the usual positive errnos is unchanged and negatives now format correctly. added a couple of negative cases (including INT_MIN) to test_strerror.


dxbjavid:
TY!

PR #943, opened 2026-07-13
fix installing json-c headers with meson and missing pkgconfig file

### LINKED ISSUE #938: Build with meson does not follow that of cmake install
Note: for general questions and comments, please use the forums at:
https://groups.google.com/g/json-c

**Describe the bug**
A clear and concise description of what the bug is, and any information about where you're running into the bug that you feel might be relevant.

Meson install does not install these files:

usr/lib/cmake/json-c/json-c-targets-release.cmake
usr/lib/cmake/json-c/json-c-config.cmake
usr/lib/cmake/json-c/json-c-targets.cmake
usr/lib/libjson-c.a
usr/lib/pkgconfig/json-c.pc
usr/include/json_pointer.h
usr/include/json_patch.h

Include files are not installed within usr/include/json-c subdirectory, but are installed in usr/include

**Steps To Reproduce**
List the steps to reproduce the behavior.
If possible, please attach a sample json file and/or a minimal code example.

Build json-c-0.19 with meson

**Version and Platform**
- json-c version: [e.g. json-c-0.14, or a specific commit hash] json-c-0.19
- OS: [e.g. Ubuntu 20.04, Debian Buster, NetBSD 9, etc...] Ubuntu 26.04
- Custom cmake/build flags, if any - None (just change of cmake to meson)

vonericsen:
Partially resolves #938

This commit fixes the install path for json-c headers when running `meson install`. Previously some showed in /usr/include/ rather than /usr/include/json-c

This also adds generating the pkgconfig .pc file for this project.

Not sure what to do about generated CMake files yet. Can discuss options on the issue instead of the PR.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -265,12 +265,37 @@ libjson = library('json-c',
 jsonc_dep = declare_dependency(link_with: libjson, include_directories: inc)
 meson.override_dependency('json-c', jsonc_dep)
 
-# Install headers
-install_headers(
+# Install headers into json-c/ subdirectory (matches CMake layout)
+installed_headers = [
   'arraylist.h', 'debug.h', 'json_c_version.h', 'json_inttypes.h',
   'json_object.h', 'json_object_iterator.h', 'json_tokener.h',
   'json_types.h', 'json_util.h', 'json_visit.h', 'linkhash.h',
   'printbuf.h', json_configure_header, json_header
+]
+
+if not get_option('disable_json_pointer')
+  installed_headers += ['json_pointer.h']
+  if not get_option('disable_json_patch')
+    installed_headers += ['json_patch.h']
+  endif
+endif
+
+install_headers(installed_headers, subdir: 'json-c')
+
+# pkg-config file
+configure_file(
+  input: 'json-c.pc.in',
+  output: 'json-c.pc',
+  install: true,
+  install_dir: get_option('libdir') / 'pkgconfig',
+  configuration: {
+    'prefix': get_option('prefix'),
+    'exec_prefix': get_option('prefix'),
+    'libdir': get_option('libdir'),
+    'includedir': get_option('includedir'),
+    'VERSION': meson.project_version(),
+    'LIBS': '',
+  }
 )
 
 # Optional apps
@@ -281,4 +306,4 @@ endif
 # Optional tests
 if get_option('buildtype') == 'debug'
   subdir('tests')
-endif
+endif 


meson.build
```diff
@@ -265,12 +265,37 @@ libjson = library('json-c',
...
+  output: 'json-c.pc',
+  install: true,
+  install_dir: get_option('libdir') / 'pkgconfig',
+  configuration: {
+    'prefix': get_option('prefix'),
+    'exec_prefix': get_option('prefix'),
+    'libdir': get_option('libdir'),
+    'includedir': get_option('includedir'),
+    'VERSION': meson.project_version(),
+    'LIBS': '',
```
eli-schwartz, line 297: This is very broken. `get_option('includedir')` returns the subpath e.g. "include/" not including "/usr". This is basic documentation at https://mesonbuild.com/Reference-manual_functions.html#get_option

> Note that the value returned for built-in options that end in `dir` such as `bindir` and `libdir` is usually a path relative to (and inside) the `prefix` but you should not rely on that, as it can also be an absolute path [in some cases](https://mesonbuild.com/Builtin-options.html#universal-options). [`install_dir` arguments](https://mesonbuild.com/Installing.html) handle that as expected but if you need an absolute path, e.g. to use in a define etc., you should use the path concatenation operator like this: `get_option('prefix') / get_option('localstatedir')`. Never manually join paths as if they were strings.

Result: `-Iinclude/json-c` per https://bugs.gentoo.org/979488 etc.

See #948 for the fix.

vonericsen: Thank you for pointing out the path issue. I did not realize that when I made this change.

Looking at #948, switching from the `configure_file` to `pkg.generate` means Windows compiles no longer get the `json-c.pc` file unless someone adds pkgconfig to Windows to build and distribute this library. That seems like an unnecessary addition to the build process.
I tested that with both MSVC compilers and Msys2, and `configure_file` always generated it for the builds I tested.
That is why I used the `configure_file` as it will always output regardless of what compiler/environment is used.

I suspect this could throw off Windows users also expecting this file similar to what was opened in #938.

I will make that change to fix the paths with the `configure_file` call and open a pull request. Would you like to review it once that is done?

eli-schwartz: > Looking at #948, switching from the `configure_file` to `pkg.generate` means Windows compiles no longer get the `json-c.pc` file unless someone adds pkgconfig to Windows to build and distribute this library. That seems like an unnecessary addition to the build process.

What???

vonericsen: I figured it out after another cup of coffee. It gets output into the `build/meson-private` directory instead of the `build` directory like the `configure_file` method I was using did when running `meson compile`.

Now that I figured that out, I do agree that the pkgconfig module is the better solution here.

eli-schwartz: Your statement implied you tested it both ways and determined that when `pkg-config` was installed, it would be generated. :) :)

Anyways the question you should be asking is what gets installed when you run `meson install --destdir=$PWD/staging_root`.

vonericsen: Yeah, that was very unclear on my part, but I'll just summarize as yesterday was not my day to work on contributing to projects...hopefully future contributions will be better :)

> Anyways the question you should be asking is what gets installed when you run meson install --destdir=$PWD/staging_root.

Yep...my laziness got the best of me, and I stopped at `meson compile`. When I revisited it to figure this all out, I did run the install which cleared it up pretty quickly.

Thanks again for taking the time to tell me what was wrong and the right way to resolve this!

Issue #944, opened 2026-07-14
NULL dereference in json_patch_apply() when JSON Patch string fields are null


zgzm78:
Summary

`json_patch_apply()` can crash when a JSON Patch operation contains `null` for fields that are expected to be strings, such as `op`, `from`, or `path`.

The function checks whether these fields exist, but does not validate their type before using the result of `json_object_get_string()` in string functions such as `strcmp()`, `strlen()`, and `strncmp()`.

## Affected commit

Tested on current `master` at commit:

```text
e65444a258435a8b52a09f182567e15c4aed5f28
```

## Impact

An application that applies untrusted JSON Patch documents through `json_patch_apply()` may be crashed by a malformed but valid JSON input.

This appears to be a potential denial-of-service issue.

## PoC

Save the following as `poc_json_patch_null_fields.c`:

```c
#include <stdio.h>
#include <json.h>
#include <json_patch.h>

static void run_poc(const char *name, const char *patch_json)
{
    struct json_object *base = json_tokener_parse("{\"a\":1}");
    struct json_object *patch = json_tokener_parse(patch_json);
    struct json_patch_error err;

    printf("[*] Running PoC: %s\n", name);
    printf("[*] Patch: %s\n", patch_json);
    fflush(stdout);

    json_patch_apply(NULL, patch, &base, &err);

    printf("[*] json_patch_apply returned without crashing\n");

    json_object_put(base);
    json_object_put(patch);
}

int main(void)
{
    run_poc("null op", "[{\"op\":null,\"path\":\"/a\"}]");
    run_poc("null from", "[{\"op\":\"move\",\"from\":null,\"path\":\"/b\"}]");
    run_poc("null path", "[{\"op\":\"move\",\"from\":\"/a\",\"path\":null}]");

    return 0;
}
```

Build example when json-c is installed system-wide:

```sh
gcc poc_json_patch_null_fields.c -ljson-c -o poc_json_patch_null_fields
```

Run:

```sh
./poc_json_patch_null_fields
```

Expected result: the process crashes before printing:

```text
[*] json_patch_apply returned without crashing
```

If building directly against the in-tree source files, one possible compile command is:

```sh
gcc -g -O0 -I . poc_json_patch_null_fields.c \
  arraylist.c debug.c json_c_version.c json_object.c json_object_iterator.c \
  json_pointer.c json_patch.c json_tokener.c json_util.c linkhash.c \
  printbuf.c random_seed.c strerror_override.c json_visit.c \
[...]

PR #946, opened 2026-07-16
reject null op/path/from fields in json_patch_apply

dxbjavid:
json_patch_apply() reads the op, path and from fields with json_object_get_string() and then hands the result straight to strcmp()/strlen(), but RFC 6902 requires those fields to be strings. If a patch document sets any of them to JSON null, which json-c represents as a NULL json_object, get_string() returns NULL and applying an untrusted patch segfaults, as reported in #944. This rejects a null op/path/from with EINVAL through the existing patch_error path before it is used, which also gives the pre-existing null-path test a clearer error message.

### DIFF
--- a/json_patch.c
+++ b/json_patch.c
@@ -209,6 +209,10 @@ static int json_patch_apply_move_copy(struct json_object **res,
 	}
 
 	from_s = json_object_get_string(jfrom);
+	if (from_s == NULL) {
+		_set_err(EINVAL, "Patch object 'from' field is not a string");
+		return -1;
+	}
 
 	from_s_len = strlen(from_s);
 	if (strncmp(from_s, path, from_s_len) == 0) {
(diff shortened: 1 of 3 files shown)


tests/json_patch_tests.json
```diff
@@ -542,6 +542,18 @@
...
+
+    { "comment": "null op field must be rejected, not dereferenced",
+       "doc": {"foo": "bar"},
+       "patch": [{"op": null, "path": "/foo"}],
+       "error": "a null op field should fail rather than crash"
+    },
+
+    { "comment": "null from field must be rejected, not dereferenced",
+       "doc": {"foo": "bar"},
+       "patch": [{"op": "move", "from": null, "path": "/foo"}],
```
hawicz, line 555: Can you please add the "path": null case too?

json_patch.c
```diff
@@ -209,6 +209,10 @@ static int json_patch_apply_move_copy(struct json_object **res,
 	}
 
 	from_s = json_object_get_string(jfrom);
+	if (from_s == NULL) {
+		_set_err(EINVAL, "Patch object 'from' field is not a string");
```
hawicz, line 213: Let's fix the get_string call so it _requires_ an actual string too. i.e.:

```
from_s = json_object_get_type(jfrom) == json_type_string ? json_object_get_string(jfrom) : NULL;
```

tests/json_patch_tests.json
dxbjavid: Added it, along with the other two. There was already a null-path case further up the file, but having the three sit together makes the intent clearer, and the message it reports is now the new one.

json_patch.c
dxbjavid: Good point, that's the stricter check. I've applied the type guard to all three fields, so a number or an object in op/path/from is now rejected the same way a null is rather than being silently stringified. The existing expected output didn't move apart from the null-path message.

PR #948, opened 2026-07-18
meson: wire up `disable_extra_libs` option

thesamesam:
This is more elegant as a meson feature option because feature options
propagate in a useful way to `
dependency(..., required: ...)`.

Without this, the option was ignored. Wiring it up naively meant that
disabling it still searched for libbsd as well, so a feature option
is the best fit.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,8 @@ conf_data = configuration_data()
 jconf_data = configuration_data()
 conf_data.set('VERSION', meson.project_version())
 
+libjson_deps = []
+
 has_std_lib = cc.has_header('stdlib.h')
 has_std_arg = cc.has_header('stdarg.h')
 has_string = cc.has_header('string.h')
@@ -20,10 +22,7 @@ if has_std_lib and has_std_arg and has_string and has_float
   conf_data.set('STDC_HEADERS', 1, description : 'Define to 1 if you have the ANSI C header files.')
 endif
 
-bsd_dep = dependency('libbsd', required: false)
-
 headers = {
-  'bsd/stdlib.h': bsd_dep,
   'dlfcn.h': [],
   'endian.h': [],
   'fcntl.h': [],
@@ -47,6 +46,15 @@ headers = {
   'xlocale.h': [],
 }
 
+may_need_libbsd = true
+if cc.has_function('arc4random')
+  may_need_libbsd = false
+  conf_data.set('HAVE_@0@'.format('arc4random'.underscorify().to_upper()), 1, description :  'Define to 1 if you have the <@0@> header file'.format('arc4random'))
+else
+  bsd_dep = dependency('libbsd', required: get_option('extra_libs'))
+  headers += { 'bsd/stdlib.h': [bsd_dep] }
+endif
+
 foreach h, d : headers
   if cc.has_header(h, dependencies: d)
     conf_data.set('HAVE_@0@'.format(h.underscorify().to_upper()), 1, description :  'Define to 1 if you have the <@0@> header file'.format(h))
@@ -63,29 +71,35 @@ if cc.has_header('stdint.h')
   jconf_data.set('JSON_C_HAVE_STDINT_H', 1, description : 'Define to 1 if you have the <stdint.h> header file.')
 endif
 
-funcs = [
-  'open',
-  'realloc',
-  'setlocale',
-  'strdup',
-  'strerror',
-  'uselocale',
-  'duplocale',
-  'vsyslog',
-  'getrandom',
-  'getrusage',
-  'strtoll',
-  'strtoull',
-  'arc4random',
-  'vasprintf',
-]
+funcs = {
+  'open': [],
+  'realloc': [],
+  'setlocale': [],
+  'strdup': [],
+  'strerror': [],
+  'uselocale': [],
+  'duplocale': [],
+  'vsyslog': [],
+  'getrandom': [],
+  'getrusage': [],
+  'strtoll': [],
+  'strtoull': [],
+  'vasprintf': [],
+}
 
-if conf_data.has('HAVE_STRINGS_H')
-  funcs += ['strcasecmp', 'strncasecmp']
+if may_need_libbsd
+  funcs += { 'arc4random': [bsd_dep] }
+  libjson_deps += [bsd_dep]
+else
+  funcs += { 'arc4random': [] }
 endif
 
-foreach f : funcs
-  if cc.has_function(f)
+if conf_data.has('HAVE_STRINGS_H')
+  funcs += {'strcasecmp': [], 'strncasecmp': []}
+endif
+
+foreach f, d : funcs
+  if cc.has_function(f, dependencies: d)
     conf_data.set('HAVE_@0@'.format(f.to_upper()), 1, description : 'Define to 1 if you have the `@0@` function.'.format(f))
   endif
 endforeach
@@ -255,7 +269,7 @@ inc = include_directories('.')
 libjson = library('json-c',
   sources,
   include_directories: inc,
-  dependencies: bsd_dep,
+  dependencies: libjson_deps,
   install: true,
   link_args: sym,
   version: '5.4.0',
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -1,5 +1,5 @@
 
-project('json-c', 'c', version: '0.18.99',
+project('json-c', 'c', version: '0.19',
```
hawicz: This should be bumped up to 0.19.99.  Also, please mention the need to update this file in RELEASE_CHECKLIST.txt (both this version and the "5.5.0" version below)

thesamesam: Will do, I also have another patch coming based on a problem we found once deployed.

meson.build
```diff
@@ -304,6 +304,6 @@ if get_option('build_apps') and host_machine.system() != 'windows'
 endif
 
 # Optional tests
-if get_option('buildtype') == 'debug'
+if get_option('build_tests')
```
eli-schwartz: Or just always define tests!

https://github.com/json-c/json-c/blob/1e39345938838a630c2d79868d81c37b531442a0/tests/meson.build#L2-L8
https://github.com/json-c/json-c/blob/1e39345938838a630c2d79868d81c37b531442a0/tests/meson.build#L67-L71

For both of these, add `build_by_default: false`. Building json-c will then never waste time compiling tests, unless one runs `meson test`. Tests are always defined, but ignored if you don't run the testsuite explicitly.

meson.build
```diff
@@ -20,7 +20,7 @@ if has_std_lib and has_std_arg and has_string and has_float
   conf_data.set('STDC_HEADERS', 1, description : 'Define to 1 if you have the ANSI C header files.')
 endif
 
-bsd_dep = dependency('libbsd', required: false)
+bsd_dep = dependency('libbsd', required: get_option('extra_libs'))
```
eli-schwartz: Slightly more complicated... it is needed for https://github.com/json-c/json-c/blob/1e39345938838a630c2d79868d81c37b531442a0/meson.build#L79 but one could do full compile/link checks for this symbol with + without libbsd to see if it is needed. At the moment it's broken I guess because https://github.com/json-c/json-c/blob/1e39345938838a630c2d79868d81c37b531442a0/meson.build#L26 handles deps for headers, but function checks don't?

It "works" ok for glibc 2.36+ since libbsd isn't needed there.

eli-schwartz (changes requested):

meson.build
thesamesam: Thanks. I need to think about the best way of laying this out.

hawicz:
Changes are looking ok so far, but something needs to be tweaked - a lot of checks are failing (no libbsd on alpine, no pkg-config on windows, etc...)

meson_options.txt
```diff
@@ -4,7 +4,7 @@ option('disable_thread_local_storage', type: 'boolean', value: false, descriptio
 option('enable_rdrand', type: 'boolean', value: false, description: 'Enable RDRAND Hardware RNG')
 option('enable_threading', type: 'boolean', value: false, description: 'Enable partial threading support')
 option('override_get_random_seed', type: 'boolean', value: false, description: 'Override json_c_get_random_seed()')
-option('disable_extra_libs', type: 'boolean', value: false, description: 'Avoid linking extra libraries like libbsd')
+option('extra_libs', type: 'feature', value: 'enabled', description: 'Allow linking extra libraries like libbsd')
```
eli-schwartz: The feature is *technically* not "extra libs" but rather "require arc4random at all costs". At any rate, making it enabled by default means CI then fails where libbsd wasn't installed (musl + alpine, Ubuntu 22.04 + old glibc, windows).

The windows issue implies that arc4random really ought not be mandatory.

hawicz: meson.build:54 has: 
```
bsd_dep = dependency('libbsd', required: get_option('extra_libs'))
```

Which makes libbsd mandatory, not arc4random, but _neither_ of those are meant to be mandatory.  I'm not all that familiar with meson syntax (is `-=` even a valid operator?), but I imagine you instead need something like:
```
if cc.has_function('arc4random')
...
else if get_option('extra_libs')
  bsd_dep = dependency('libbsd', required: false)
  if bsd_dep
      headers += { 'bsd/stdlib.h': [bsd_dep] }
      if cc.has_function('arc4random')
        conf_data.set(...)
      else
        headers -= { 'bsd/stdlib.h': [bsd_dep] }
      endif
  endif
endif
```

hawicz (changes requested):

meson_options.txt
eli-schwartz: > meson.build:54 has:
> 
> ```
> bsd_dep = dependency('libbsd', required: get_option('extra_libs'))
> ```
> 
> Which makes libbsd mandatory, not arc4random, but _neither_ of those are meant to be mandatory.

It happens:
- inside of an if block, so libbsd is never mandatory unless arc4random is not found, hence this is in fact all about arc4random
- via Meson's builtin "features" option type, so `required: get_option('extra_libs')` can be defined on the command line as -Dextra_libs=auto, in which case it is not mandatory since it will be "not-found" and simply say "ok, arc4random cannot be found even with libbsd? Fine, disable the arc4random define". Similarly, users could pass -Dextra_libs=disabled on the command line and it will be force disabled, meson won't even look for libbsd... But it does look for arc4random via "required: false" so there's no option to turn it off if it comes with the platform standard C library. :)

The fix here can simply be to make it not-required by default:

```suggestion
option('extra_libs', type: 'feature', value: 'auto', description: 'Allow linking extra libraries like libbsd')
```

since the whole purpose of "enabled" is to say "force toggled on, error out if not found".


> I'm not all that familiar with meson syntax (is `-=` even a valid operator?), but I imagine you instead need something like:
> 
> ```
> if cc.has_function('arc4random')
> ...
> else if get_option('extra_libs')
>   bsd_dep = dependency('libbsd', required: false)
>   if bsd_dep
>       headers += { 'bsd/stdlib.h': [bsd_dep] }
>       if cc.has_function('arc4random')
>         conf_data.set(...)
>       else
>         headers -= { 'bsd/stdlib.h': [bsd_dep] }
>       endif
>   endif
> endif
> ```

You'd simply omit it by default and add it when wanted, there is absolutely zero purpose in having the if branch of a conditional add something to a dictionary and the else branch remove it. If it was there to be removed, the if branch wouldn't add it. If it wasn't there to be removed, the else branch doesn't have anything to remove and would fatally error out with a key lookup exception.

hawicz: > no option to turn it off if it comes with the platform standard C library. 

If arc4random is available, in whatever library, then it's intended to be used.  Addition some new config option to skip using arc4random is a different topic entirely.

> so libbsd is never mandatory unless arc4random is not found

That's broken.  libbsd should be included only if that's where arc4random actually exists.

The aim of these checks, at least in the cmake build, is:
Use arc4random from whatever library it happens to be, by default including "extra libraries" (i.e. libbsd, but also any other library that we may later realize needs to be checked) .  If we've been told _not_ to look in extra libraries (i.e. `DISABLE_EXTRA_LIBS`) then only check for arc4random in the libraries that we build with by default.

eli-schwartz: > That's broken. libbsd should be included only if that's where arc4random actually exists.

I'm not sure where I went wrong then, because I thought that's exactly what I said this PR does.

hawicz: I had a chance to sit down and figure out some meson syntax and take another look at this change.  The problem with it is that it's checking for arc4random in two different places.  Here's a tweak that eliminates the extra initial check, and makes the later has_function() loop a bit smarter: https://github.com/hawicz/json-c/commit/27bf6592043dec8d237f830f20e2ed515d1cadd6

Assuming neither of you see anything wrong with that, @thesamesam please fetch this into your fork/branch, and then I'll merge this PR.

thesamesam: Will check it out tonight or tomorrow, sorry, busy week

thesamesam: OK, thanks. I think your change is fine and I agree it's more elegant. I think the PR as it is would've worked with the enabled->auto change but I prefer the deduplication in yours anyway.

PR #949, opened 2026-07-20
reject empty reference token as array index in json_pointer

dxbjavid:
is_valid_index treats a zero-length reference token as valid and hands back index 0, since the empty string skips the single-digit fast path, is not caught by the leading-zero check, and then falls through to strtoull on an empty string. The practical effect is that a pointer like "/" or "/a/" against an array silently resolves to element 0 rather than failing, and because json_patch_apply goes through the same helper an RFC 6902 document with a path of "/" will replace or remove the first element instead of being rejected. Rejecting the empty token with EINVAL at the top of is_valid_index covers both the get and set call sites; empty-string object keys are unaffected, as the helper is only consulted for arrays.

### DIFF
--- a/json_pointer.c
+++ b/json_pointer.c
@@ -46,6 +46,12 @@ static void string_replace_all_occurrences_with_char(char *s, const char *occur,
 static int is_valid_index(const char *path, size_t *idx)
 {
 	size_t i, len = strlen(path);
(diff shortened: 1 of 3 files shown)


dxbjavid:
Thanks.

PR #950, opened 2026-07-21
bug: Fixing pkgconfig paths and math dependency

vonericsen:
Fixing the paths in the pkgconfig .pc file since I did not realize in json-c/json-c/#943 that these variables use relative paths. Adding getoption('prefix') / to concatenate them resolves this.

Also noticed when comparing the output to Cmake's generated .pc file that the math dependency (-lm) was missing in Linux builds, so added that in as well.

The path issue was pointed out in a comment in #943 by @eli-schwartz


vonericsen:
Cancelling as this is not the correct solution

PR #951, opened 2026-07-22
Proper pkgconfig with meson

vonericsen:
Per my discussion with @eli-schwartz on #943, I implemented the generation of the `json-c.pc` file incorrectly.

This is the correct way to generate it using the pkgconfig module in meson which handles this in a much more correct and simpler way.

When comparing to the CMake generated output, I also noticed a missing `-lm` dependency in the `json-c.pc` file so this also corrects that error/inconsistency.

This method is part of what is in progress on #948 as well, however that draft is not completed and I wanted to get this correct solution in place ASAP.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 
-project('json-c', 'c', version: '0.18.99',
+project('json-c', 'c', version: '0.19.99',
   meson_version: '>=0.54.0',
   license: 'MIT',
   default_options: ['buildtype=release', 'warning_level=2'])
@@ -251,11 +251,14 @@ endif
 # Include directories
 inc = include_directories('.')
 
+# Math library (needed for INFINITY, isnan, etc.)
+m_dep = cc.find_library('m', required: false)
+
 # Build library
 libjson = library('json-c',
   sources,
   include_directories: inc,
-  dependencies: bsd_dep,
+  dependencies: [bsd_dep, m_dep],
   install: true,
   link_args: sym,
   version: '5.4.0',
@@ -287,6 +290,7 @@ pkg = import('pkgconfig')
 pkg.generate(
   libjson,
   description: 'A JSON implementation in C',
+  libraries_private: m_dep,
   subdirs: 'json-c',
 )
 
--- a/meson.build
+++ b/meson.build
@@ -265,7 +265,7 @@ libjson = library('json-c',
 jsonc_dep = declare_dependency(link_with: libjson, include_directories: inc)
 meson.override_dependency('json-c', jsonc_dep)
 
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> This method is part of what is in progress on #948 as well, however that draft is not completed and I wanted to get this correct solution in place ASAP.

Then why not use `git cherry-pick` on that draft PR? It's only a draft because of an unrelated change that @thesamesam is still fine-tuning... The existing commit that someone else already made to fix the pkgconfig file is, itself, not a draft.

meson.build
```diff
@@ -251,11 +251,14 @@ endif
 # Include directories
 inc = include_directories('.')
 
+# Math library (needed for INFINITY, isnan, etc.)
+m_dep = cc.find_library('m', required: false)
```
eli-schwartz, line 255: Indeed, this isn't strictly related to pkgconfig at all. It's actually just an issue whereby POSIX guarantees that math.h provides functions which "shall be available to the software if linked with -lm".

Some operating systems including with recent versions of glibc, may install a stub libm.a because their libc.so already provides all math symbols. So testing won't necessarily notice that this was missing.

Meson documents this here: https://mesonbuild.com/howtox.html#add-math-library-lm-portably

meson.build
```diff
@@ -287,6 +290,7 @@ pkg = import('pkgconfig')
 pkg.generate(
   libjson,
   description: 'A JSON implementation in C',
+  libraries_private: m_dep,
```
eli-schwartz, line 293: ... Why do you need this though? Meson's pkgconfig generator is interesting precisely because it normally handles recursive dependencies as "private libraries" automatically for you. If it's a dependency of the library itself, that should suffice.

vonericsen: I was comparing the output file from meson to the cmake build system which was listing this.
It seemed like something that was added for some other reason and would be needed here as well, but thinking about this today and what I did see in the meson documentation, I do understand why this is not needed now.

vonericsen:
> Then why not use git cherry-pick on that draft PR? It's only a draft because of an unrelated change that @thesamesam is still fine-tuning... The existing commit that someone else already made to fix the pkgconfig file is, itself, not a draft.

I did not realize I could do that across forks/remotes until now.
I just tried it locally and now know how to do that.

I apologize for not realizing that was possible. I will watch the other PR to see what else I can learn instead of keeping this one open.

thesamesam:
No worries @vonericsen and thanks a lot for caring about getting it sorted! I'll cherry-pick your `libm` fix.

eli-schwartz:
Yup, the libm fix is a nice find and deserves to be landed (and credited). 

@thesamesam that being said, note my review comment above regarding pkg.generate.

PR #952, opened 2026-07-22
Implementing extra_libs for linux

vonericsen:
This PR improves Meson handling of extra_libs for arc4random detection and fallback behavior, building on the work discussed in #948.

# What changed

- Replaced disable_extra_libs (boolean) with extra_libs (feature).
- Wired the option into detection/link behavior (it previously did not control the fallback path).
- Switched to capability-first detection for arc4random:

1. Check for arc4random in system libc first.
2. If not found, and extra_libs is enabled, check bsd/stdlib.h + libbsd.

- Made libbsd usage conditional on actually taking the fallback path, so it is not linked/exported unnecessarily.
- Kept pkg-config output aligned with link behavior by only including fallback private libs when used.

# Why

Different platforms/`libc`s provide arc4random differently:

- BSD libc: typically provides it directly.
- Newer glibc: may provide it directly.
- Some Linux/musl environments: may require fallback behavior.
- This ordering preserves native libc behavior when available and only uses libbsd as an opt-in fallback.

# Behavior impact

- If libc provides arc4random, it is used.
- If libc does not provide it, fallback to libbsd is attempted only when extra_libs is enabled.
- If fallback is not available, existing non-arc4random seed paths remain in place.

# Validation
Tested with:

- glibc environment where arc4random is available in libc.
- musl environment where arc4random is not available in libc.
- BSD targets continue to work correctly due to libc-first detection order.

# Context

Related draft: #948
This approach aims to keep the same goal while making the detection and dependency behavior more explicit and conditional.

@eli-schwartz @thesamesam, does this direction look better than the draft approach in #948?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -20,7 +20,11 @@ if has_std_lib and has_std_arg and has_string and has_float
   conf_data.set('STDC_HEADERS', 1, description : 'Define to 1 if you have the ANSI C header files.')
 endif
 
-bsd_dep = dependency('libbsd', required: false)
+extra_libs = get_option('extra_libs')
+bsd_dep = disabler()
+if not extra_libs.disabled()
+  bsd_dep = dependency('libbsd', required: false)
+endif
 
 headers = {
   'bsd/stdlib.h': bsd_dep,
@@ -76,7 +80,6 @@ funcs = [
   'getrusage',
   'strtoll',
   'strtoull',
-  'arc4random',
   'vasprintf',
 ]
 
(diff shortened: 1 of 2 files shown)


eli-schwartz (changes requested):
> This approach aims to keep the same goal while making the detection and dependency behavior more explicit and conditional.
> 
> @eli-schwartz @thesamesam, does this direction look better than the draft approach in #948?

This direction is semantically wrong in ways that look extremely verbose and use design patterns one normally wouldn't expect to reach for in such a case. I somehow suspect that it was LLM assisted as a result...

... which is normally my cue to disengage. Sorry.

vonericsen:
I used a LLM to help me format the message of the PR after I made the changes because I was not satisfied with what I was writing myself....it never seemed like my thoughts were clear.

I will watch the other PR to see what I can learn for the future.

PR #955, opened 2026-08-01
printbuf: preserve self-append sources across growth

carrerasdarren-cell:
## Summary

- preserve internal source pointers when `printbuf_memappend()` grows the buffer
- use `memmove()` for an aliased copy
- cover self-append across a resize boundary

## Problem

When `buf` points into `p->buf`, `printbuf_extend()` may move the allocation and invalidate `buf` before the copy. The subsequent `memcpy()` then reads from the freed allocation. A straightforward example is appending the current printbuf contents to itself when the combined contents require growth.

The fix records an internal source's offset before `realloc()` and rebases it onto the resulting allocation.

## Validation

- GCC 14 with ASan and UBSan: self-append PoC passes while `realloc()` moves the buffer
- GCC 14 release build with `-Wall -Wextra -Werror`: 28/28 tests pass
- AppleClang 17 release build with `-Wall -Wextra -Werror`: printbuf test and all 27 portable tests pass

### DIFF
--- a/printbuf.c
+++ b/printbuf.c
@@ -28,6 +28,7 @@
 #endif /* HAVE_STDARG_H */
 
 #include "debug.h"
+#include "json_inttypes.h"
 #include "printbuf.h"
 #include "snprintf_compat.h"
 #include "vasprintf_compat.h"
@@ -96,6 +97,8 @@ static int printbuf_extend(struct printbuf *p, int min_size)
 
 int printbuf_memappend(struct printbuf *p, const char *buf, int size)
 {
+	int buf_offset = -1;
+
 	/* Prevent signed integer overflows with large buffers. */
 	if (size < 0 || size > INT_MAX - p->bpos - 1)
 	{
@@ -104,10 +107,18 @@ int printbuf_memappend(struct printbuf *p, const char *buf, int size)
 	}
 	if (p->size <= p->bpos + size + 1)
 	{
+		const uintptr_t buf_addr = (uintptr_t)(const void *)buf;
+		const uintptr_t p_buf_addr = (uintptr_t)(const void *)p->buf;
+
+		/* realloc() invalidates source pointers into the old buffer. */
+		if (buf_addr >= p_buf_addr && buf_addr - p_buf_addr < (uintptr_t)p->size)
+			buf_offset = (int)(buf_addr - p_buf_addr);
 		if (printbuf_extend(p, p->bpos + size + 1) < 0)
 			return -1;
+		if (buf_offset >= 0)
+			buf = p->buf + buf_offset;
 	}
-	memcpy(p->buf + p->bpos, buf, size);
+	memmove(p->buf + p->bpos, buf, size);
 	p->bpos += size;
 	p->buf[p->bpos] = '\0';
 	return size;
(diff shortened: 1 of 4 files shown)


tests/test_printbuf.c
```diff
@@ -126,6 +126,38 @@ static void test_printbuf_memappend(int *before_resize)
...
+	assert(printbuf_memappend(pb, data, data_size) == data_size);
+	free(data);
+	assert(pb->size == initial_size);
+
+	assert(printbuf_memappend(pb, pb->buf, pb->bpos) == data_size);
+	assert(pb->size > initial_size);
+	assert(pb->bpos == data_size * 2);
+	for (i = 0; i < pb->bpos; i++)
+		assert(pb->buf[i] == 'X');
+	assert(pb->buf[pb->bpos] == '\0');
```
hawicz, line 155: Your fix _also_ fixes things like `sprintf(p, "%s", some_location_inside_p_buf)`, would you mind extending this test to check that as well?

printbuf.c
```diff
@@ -104,10 +107,18 @@ int printbuf_memappend(struct printbuf *p, const char *buf, int size)
 	}
 	if (p->size <= p->bpos + size + 1)
 	{
+		const uintptr_t buf_addr = (uintptr_t)(const void *)buf;
```
hawicz, line 109: Let's move the declaration of `buf_offset` down into this block.

printbuf.c
```diff
@@ -104,10 +107,18 @@ int printbuf_memappend(struct printbuf *p, const char *buf, int size)
...
+		/* realloc() invalidates source pointers into the old buffer. */
+		if (buf_addr >= p_buf_addr && buf_addr - p_buf_addr < (uintptr_t)p->size)
+			buf_offset = (int)(buf_addr - p_buf_addr);
 		if (printbuf_extend(p, p->bpos + size + 1) < 0)
 			return -1;
+		if (buf_offset >= 0)
+			buf = p->buf + buf_offset;
 	}
-	memcpy(p->buf + p->bpos, buf, size);
+	memmove(p->buf + p->bpos, buf, size);
```
hawicz, line 120: Since this is overlap-safe now, can you please update the doc comments in printbuf.h to mention that?  (both for `printbuf_memappend()` and `sprintfbuf()`)

tests/test_printbuf.c
carrerasdarren-cell: Added a growth-forcing `sprintbuf(pb, "%s", pb->buf + 1)` regression in 9782370. It exercises an interior source pointer, forces reallocation, and verifies the complete result.

printbuf.c
carrerasdarren-cell: Moved `buf_offset` into the growth block in 9782370.

printbuf.c
carrerasdarren-cell: Updated both public API comments in 9782370 to document aliased/overlapping `printbuf_memappend()` sources and `sprintbuf()` arguments that point into the print buffer.

hawicz (approved):

hawicz:
Thanks for the PR!

PR #958, opened 2026-08-17
reject negative after whitespace in json_parse_uint64

dxbjavid:
json_parse_uint64 only skips a leading space before rejecting a '-', but strtoull skips the whole whitespace set and then quietly negates the sign, so a value like "\t-1" (or one behind a newline, carriage return, form feed or vertical tab) slips past the guard and comes back as a wrapped near-UINT64_MAX value instead of an error. The same path is reachable from a JSON string through json_object_get_uint64, so a document controlling that string can turn a negative into a huge unsigned result. This skips the same whitespace strtoull does so the existing '-' rejection actually holds, with a regression case for the tab form added to test_parse_int64.

### DIFF
--- a/json_util.c
+++ b/json_util.c
@@ -263,7 +263,13 @@ int json_parse_uint64(const char *buf, uint64_t *retval)
 	uint64_t val;
 
 	errno = 0;
-	while (*buf == ' ')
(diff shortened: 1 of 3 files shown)


dxbjavid:
Thanks

